PR #2, opened 2017-05-30
add deserializeUnknown method

### LINKED ISSUE #1: QueryParams backward compatibility
Is it somehow possible to support `group` in QUeryParams.  
My current project makes heavy use of this old pattern.  
  
Currently accessing an endpoint with `group` produces 
```Java
13:31:56,249 ERROR [io.crnk.core.engine.internal.dispatcher.HttpRequestProcessorImpl] (default task-3) failed to process request: java.lang.IllegalStateException: group
	at io.crnk.core.queryspec.DefaultQuerySpecDeserializer.deserialize(DefaultQuerySpecDeserializer.java:172)
	at io.crnk.core.queryspec.internal.QuerySpecAdapterBuilder.build(QuerySpecAdapterBuilder.java:41)
	at io.crnk.core.engine.internal.dispatcher.HttpRequestProcessorImpl.dispatchRequest(HttpRequestProcessorImpl.java:109)
	at io.crnk.core.engine.internal.http.JsonApiRequestProcessor.process(JsonApiRequestProcessor.java:81)
	at io.crnk.core.engine.internal.dispatcher.HttpRequestProcessorImpl.process(HttpRequestProcessorImpl.java:78)
	at io.crnk.rs.CrnkFilter.filter(CrnkFilter.java:48)
...
```

mohlek:
fix #1
If an unknown Parameter is passed, `deserializeUnknown` will be called so you can easily intercept it.
By default it will throw `IllegalStateException` as before.

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/queryspec/DefaultQuerySpecDeserializer.java
+++ b/crnk-core/src/main/java/io/crnk/core/queryspec/DefaultQuerySpecDeserializer.java
@@ -169,7 +169,7 @@ public class DefaultQuerySpecDeserializer implements QuerySpecDeserializer {
 					deserializePage(querySpec, parameter);
 					break;
 				default:
-					throw new IllegalStateException(parameter.paramType.toString());
+					deserializeUnknown(querySpec, parameter);
 			}
 
 		}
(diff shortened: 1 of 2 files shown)


mohlek:
Im not sure but i don't think its my fault that the test fails

remmeier:
no it is not, jsut the first PR :-( 

potentially 
```
QuerySpec rootQuerySpec = new QuerySpec(resourceInformation.getResourceClass());
```

could also be moved into a new method to allow to create a subtype of QuerySpec that also carries the group?

mohlek:
are you okay with `createQuerySpec`? I really don't know how to name this method :)

remmeier:
yes, perfect

remmeier:
have to check what is happening with NEXUS_DEPLOY_PASS, because in Travis it is defined and should work. Either way the check could also be removed for branch builds as there is no publishing.

mohlek:
Remove it if its not needed.

remmeier:
I removed the check in the master branch, cannot commit that here, so the remote branch needs to be updated. Not quite sure yet what is happening, as the other PR seems to work fine.

Issue #14, opened 2017-06-08
Support paging by providing hasNext instead of total count


remmeier:
usually it is more efficient to just check whether more resources are available rather than count the total number of resources. This is sufficient to compute all except the "last" page link. An apppropriate interface should be provided and implemented by the JPA repositories.

PR #15, opened 2017-06-08
Support paging by providing hasNext instead of total count #14

remmeier:
- introduced HasMoreResourcesMetaInformation
- implemented JpaModule.setTotalResourceCountUsed
- update page link computation accordingly


remmeier:
yes sounds good to me, it currently is aligned with the company policies where I work for. But I'm not particular fond of that else on a new line...

PR #37, opened 2017-06-21
WIP: use value instead of type

### LINKED ISSUE #36: JsonApiResource "refactor"
I'd like to change `JsonApiResource` to use value and not `type`
This would allow us to use it in a cleaner way.

```java
@JsonApiResource("abc");
```

instead of:

```java
@JsonApiResource( type = "abc");
```

Having to explicit declare `type` is not necessary in my opinion since there are no other fields in the annotation.

mohlek:
just a suggestion...

fixes #36

### DIFF
--- a/crnk-brave/src/test/java/io/crnk/brave/mock/models/Project.java
+++ b/crnk-brave/src/test/java/io/crnk/brave/mock/models/Project.java
@@ -7,7 +7,7 @@ import io.crnk.core.resource.annotations.JsonApiToMany;
 import javax.validation.constraints.NotNull;
 import java.util.List;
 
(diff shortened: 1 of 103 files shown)


remmeier:
let us check use cases for more attributes in @JsonApiResource. Depending on that we can consider this for a 2.0 version resp. allowing the use of both in parallel. Currently it is a major breaking change that should be avoided.

mohlek:
I agree. This was more an example of how I think it should be used. 
The tests are failing on travis but not local. I need to investigate this anyways.
Lets close this pr for the moment and discuss about it somewhere else

PR #38, opened 2017-06-22
throw ResourceNotFoundException

### LINKED ISSUE #19: findOne is not returning 404 when entity does not exist
from @phuo1032:

This issue seemed to be fixed at some point from a PR

#272

But the latest version does not have this change any more.
https://github.com/katharsis-project/katharsis-framework/blob/master/katharsis-core/src/main/java/io/katharsis/core/internal/dispatcher/controller/ResourceGet.java

The test case that verifies the NOT_FOUND status is also not there.
https://github.com/katharsis-project/katharsis-framework/blob/master/katharsis-core/src/test/java/io/katharsis/core/internal/dispatcher/controller/resource/ResourceGetTest.java

Is this a design change for restoring the previous behavior or am I suppose to throw the NOT_FOUND exception manually?

mohlek:
fix #19  

Todo:
- [x] Throw Exception when resource is `null`
- [x] Check for property
- [x] Add test

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
+++ b/crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
@@ -1,7 +1,5 @@
 package io.crnk.core.boot;
 
-import java.util.List;
-
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import io.crnk.core.engine.error.JsonApiExceptionMapper;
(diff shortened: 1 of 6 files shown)


remmeier:
the json api specification actually allows to return null in case of a missing record: see "404 Not Found" => http://jsonapi.org/format/. And so it is up to the repository to decide whether to throw an exception or null.

But agree that people many times want the 404. So maybe this commit could be complemented with a feature flag like CrnkProperties.RETURN_404_ON_NULL

mohlek:
Yeah we should make it configurable. I'll change it.

mohlek:
I added `PropertyProvider` to `ModuleRegistry` I thought this can be handy in many ways and allows me to get the property very easy

mohlek:
@remmeier please take a look. Any suggestions?

PR #46, opened 2017-06-28
add generic cast

### LINKED ISSUE #45: Enhancement: FilterSpec generic cast getValue
I'd like to generic cast the result of `getValue`.
I don't see any backward compatibility problems with this as it was `Object`.

Using generic cast would reduce the struggle of manual casting.
If you are unsure of the return type use `Object` otherwise the return type is the same as defined in the resource-class ( thanks for this :smile: ).

mohlek:
fix #45


remmeier:
that is useful :-)

PR #53, opened 2017-07-05
set preset for null values

### LINKED ISSUE #49: Wrong response when relation is null
When using a lazy relationship and the relationship repository returns null, crnk returns a wrong response. The response should be equal to a normal resource being "null". 
Currently crnk responses with `{}` which is illegal related to the json api specs.

mohlek:
fix #49 

Checklist:
- [x] Analyse / Fix Problem
- [ ] Write a Test

The problem with the response beeing `{}` seems to be the value `present`.

```Java
public class Document implements MetaContainer, LinksContainer {

	@JsonInclude(Include.NON_EMPTY)
	@JsonDeserialize(using = DocumentDataDeserializer.class)
	@JsonSerialize(using = NullableSerializer.class)
	private Nullable<Object> data = Nullable.empty();
//...
```

data.preset is by default false.

```Java
	private void setResponse(HttpRequestContext requestContext, Response crnkResponse)
			throws IOException {
		if (crnkResponse != null) {
			ObjectMapper objectMapper = moduleContext.getObjectMapper();
			String responseBody = objectMapper.writeValueAsString(crnkResponse.getDocument());

			requestContext.setResponse(crnkResponse.getHttpStatus(), responseBody);
			requestContext.setResponseHeader("Content-Type", HttpHeaders.JSONAPI_CONTENT_TYPE_AND_CHARSET);
		}
	}
```

With present = false, `objectMapper.writeValueAsString(crnkResponse.getDocument());` will produce `{}`

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/engine/internal/dispatcher/controller/FieldResourceGet.java
+++ b/crnk-core/src/main/java/io/crnk/core/engine/internal/dispatcher/controller/FieldResourceGet.java
@@ -17,6 +17,7 @@ import io.crnk.core.engine.registry.RegistryEntry;
 import io.crnk.core.engine.registry.ResourceRegistry;
 import io.crnk.core.exception.ResourceFieldNotFoundException;
 import io.crnk.core.repository.response.JsonApiResponse;
+import io.crnk.core.utils.Nullable;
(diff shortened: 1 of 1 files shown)


remmeier:
merged this one, never the less still various things open in this area that would require a bit of cleanup

PR #60, opened 2017-07-10
JsonApiResponseFilter improvements

gianinbasler:
Checking if caller requested JSON-API response when wrapping null responses.
Also wrap simple responses (String, etc.) if caller requests response in JSON-API format.
Added a repository method returning null to ScheduleRepository.
Added tests for new functionality and adapted existing failing tests.


crnk-rs/src/main/java/io/crnk/rs/JsonApiResponseFilter.java
```diff
@@ -92,4 +101,34 @@ private boolean isResourceResponse(Object response) {
...
+	 * requested a response in JSON-API format.
+	 *
+	 * @param requestContext the current container request context as received by this filter
+	 * @return <code>true</code>, if the caller requested a JSON-API response,<br />
+	 *     		<code>false</code>, otherwise
+	 */
+	private boolean isJsonApiRequest(ContainerRequestContext requestContext) {
+		HttpRequestContextBase httpRequestContextBase = new JaxrsRequestContext(requestContext, feature);
+		HttpRequestContextBaseAdapter httpRequestContext = new HttpRequestContextBaseAdapter(httpRequestContextBase);
+		return JsonApiRequestProcessor.isJsonApiRequest(httpRequestContext);
```
remmeier: I think it should not be requested dependant. Because it may confuse developers as as they will see different results depending on whether they look at the services with a browser or do a proper json api ajax call. 

I guess there is a possiblity to specify a the content type in the jaxrs method? and maybe a default could be configured to do either all json-api and specify the exceptions or the other way around.

gianinbasler: You can specify the content type of a jaxrs method using `@Produces("application/vnd.api+json")`. This can then be queried via `ContainerResponseContext`'s `getMediaType()` method.

So, it is easiest to just check if the response context's media type is json-api and wrap the response if this is not the case yet. This way, no default has to be specified and the developer can simply specify if the response should be wrapped by adding a `@Produces` annotation.

Do you agree?

PR #102, opened 2017-09-11
Relationship repository lookup behavior overrides the global setting

csetera:
As discussed via Gitter, this PR changes the handling of the lookup behavior so that relationship annotations can override the default global setting.  It introduces a new DEFAULT for lookup behavior and makes that the actual default if no other value is specified by the relationship annotation.  When the relationship lookup behavior is set to DEFAULT, it will look to the globally set value.  If the relationship lookup behavior is set to DEFAULT and the globally value is unset, the default will be up to the resource information provider in use (NONE, usually... although JPA does something different).  If the relationship lookup behavior is set to something other than DEFAULT, then that value will be used no matter what the value of the global behavior.

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/IncludeLookupUtil.java
+++ b/crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/IncludeLookupUtil.java
@@ -32,23 +32,28 @@ public class IncludeLookupUtil {
 	}
 
 
-	public static LookupIncludeBehavior getDefaultLookupIncludeBehavior(PropertiesProvider propertiesProvider) {
+	public static LookupIncludeBehavior getGlolbalLookupIncludeBehavior(PropertiesProvider propertiesProvider) {
 		if (propertiesProvider == null) {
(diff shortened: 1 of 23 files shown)


crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/IncludeLookupUtil.java
```diff
@@ -32,23 +32,28 @@ public IncludeLookupUtil(ResourceRegistry resourceRegistry, IncludeBehavior incl
 	}
 
 
-	public static LookupIncludeBehavior getDefaultLookupIncludeBehavior(PropertiesProvider propertiesProvider) {
+	public static LookupIncludeBehavior getGlolbalLookupIncludeBehavior(PropertiesProvider propertiesProvider) {
```
remmeier: I gonna move that method somewhere else eventually

csetera:
Glad to see this merged.  Any estimates on when you might create a new release that contains this change?

remmeier:
next one, two days. I have to have a quick look at crnk-jpa.

PR #107, opened 2017-09-15
Set the PropertiesProvider on to the ModuleRegistry at the beginning of boot

csetera:
This PR changes the ordering of CrnkBoot bootstrapping to move setting of the PropertiesProvider on to the ModuleRegistry much earlier in the boot process.  With my last PR, I added the ability to retrieve the PropertiesProvider from within the ModuleContext where it could be used as necessary.  However, I didn't realize that the PropertiesProvider was not being set on the ModuleRegistry until the end of bootstrap.  This change allows for Module implementations to get the correct PropertiesProvider instance while they are in the midst of bootstrapping and fixes our use case of this functionality.

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
+++ b/crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
@@ -148,6 +148,10 @@ public class CrnkBoot {
 		checkNotConfiguredYet();
 		configured = true;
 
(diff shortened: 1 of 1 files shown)


remmeier:
thx,

moduleContext and dependencies is sometimes a bit tricky. Here still simple. At some point modulecontext may get deprecated in favor of ModuleExtension. But that is way of.

Issue #133, opened 2017-10-26
[crnk-jpa] JsonApiResource is being ignored


mohlek:
crnk version: 2.2.20171015181129

```Java
@Entity
@Table(name = "my_table")
@Getter
@Setter
@JsonApiResource(type = "awesome-class")
public class AwesomeClass {
...
}
```

Given that class, the repository should be `api/awsome-class` insead of `api/awesomeClass`

remmeier:
support needs to be added in JpaResourceInformation.getResourceType. Maybe some more flexbility in the area would be great next to support for JsonApiResource. Like being able to provide a method that does the entity to resource name mapping. Or to add a prefix like "entity/awesome-class" to all entity resources.

remmeier:
resolved in the meantime

PR #134, opened 2017-10-26
use JsonApiResource annotation

mohlek:
Quick and easy fix for #133


mohlek:
Sorry for the import mess. IntelliJ autoformatted it. If it bother you I'll revert it.

PR #135, opened 2017-10-26
WIP: use JsonApiResource annotation

mohlek:
Quick and easy fix for #133.

Sorry I fu**ed my last pr


remmeier:
thx! could you add a tiny test to JpaResourceInformationProviderTest to make sure support for this stays this way? :-) e.g. by adding the JsonApiResource annotation to AnnotationTestEntity and then check it in that test class.

mohlek:
interesting, it works on my side.

remmeier:
I know what is going one. It is a bit annoying. The JsonApiResource annotation is causing issues, as there are two ResourceInformationBuilder that are "able" to process the class. But actually only the JPA one is correct. So test will currently fail if the wrong one is chosen. Not quite sure about the solution. That @JpaResource is one, but I don't like that. Same annoations for everything is better. Currently the best solution that I did come up is to introduce a priority for the ResourceInformationBuilder with the Prioritizable interface to make sure Jpa is checked first.

PR #160, opened 2017-11-20
Added methods

vicmosin:
@remmeier Don't know whether it's a provate case or  more generic.. But in my case it would be nice to have them


remmeier:
add some point it might be more worthwhile to have a builder for all the permutations

remmeier:
thx

PR #162, opened 2017-11-21
Resolved #161

### LINKED ISSUE #161: QuerySpec's custom parameters
I couldn't find any way to support custom parameters.. All interfaces depend on `QuerySpec` which is class i.e. one can't just use inheritance or implement his own implementation (if it was an interface for example).
From other side, there is a `QuerySpecDeserializer` interface which might be overriden.. So question is what's the point of implementing your own deserializer if all of the repos are hardcoded to work with `QuerySpec`??

What if we introduce a list of unsupported (and provided) query params as  map or something.. This way one can override deserializer and have his custom fields accessable..

vicmosin:


vicmosin:
@remmeier Could you please review it? We are kinda need this feature ASAP

PR #164, opened 2017-11-22
Improve validation support

vicmosin:
Provides an optional validation in case of validation module enabled.


remmeier:
something in this direction would be greatly useful. One can achieve the same thing without hooking validation into crnk-core (to keep the it small and modular). There are various filters, see:

http://www.crnk.io/documentation/#_module_development

in this case a RepositoryFilter  should do the trick. It can intercept the requests and do the validation before forwarding to the actual repository.

vicmosin:
@remmeier Yeap, repository filter fits better, thanks. Updated.

crnk-validation/src/main/java/io/crnk/validation/ValidationModule.java
```diff
@@ -31,7 +32,7 @@ public String getModuleName() {
 	public void setupModule(ModuleContext context) {
 		context.addExceptionMapper(new ConstraintViolationExceptionMapper(context));
 		context.addExceptionMapper(new ValidationExceptionMapper());
-	}
-
 
+		context.addRepositoryFilter(new ValidationRepositoryFilter());
```
remmeier: looks great, one last minor thing, could add a boolean property to the module whether to enable/disable the filter to allow backward compatibility?

depending on the use cases, the repositories may want to execute the validator by itself. But default true like it is now seems good.

remmeier:
thx!

vicmosin:
@remmeier could you please provide release

remmeier:
is wedneyday or ok? then I would do more official larger one. A need a day or so finish the Spring topics.

vicmosin:
@remmeier yeah, can wait..

vicmosin:
@remmeier Any news on release?

remmeier:
sorry, was a busy week and hold up with other things. Sometime today after spring boot is merged. your other PR is also ready?

vicmosin:
Yes, I fixed merge conflicts , but please review it once again since you said you need to think over the whole spring boot config stuff.. What I meant with that PR is to simplify the process of injecting custom implimentations of query deserializer. Now I had to override the whole SpringBooV3 class.

PR #165, opened 2017-11-23
Added the more comfortable way to injectQuerySpecDeserializer impleme…

vicmosin:
Provides an improved version of `CrnkBoot` disables explicit `DefaultQuerySpecDeserializer` creation by adding optional injection of external implementation while using default impl as a fallback


remmeier:
looks good, I think I have to sleep over the general approach, all the properties, the matching objects (like the spring boot configuration). 

potentially also align QuerySpecSerializer and QuerySpecDeserializer a bit.

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
```diff
@@ -417,10 +439,13 @@ public void setServiceDiscovery(ServiceDiscovery serviceDiscovery) {
 	 * NOTE: This using this feature requires a {@link QuerySpecDeserializer} and it does not work with the
 	 * deprecated {@link QueryParamsBuilder}.
 	 */
+	@Deprecated
```
remmeier: I know a number of application making heacy use of this method rather than the properties (or rather CrnkFeature.set...()). . At some point in would be good to align those to mechanisms (like in Spring with the Properties that can be injected into objects). But till that point it cannot be deprecated.

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
```diff
@@ -417,10 +439,13 @@ public void setServiceDiscovery(ServiceDiscovery serviceDiscovery) {
...
 	public void setDefaultPageLimit(Long defaultPageLimit) {
 		PreconditionUtil.assertNotNull("Setting the default page limit requires using the QuerySpecDeserializer, but " +
 				"it is null. Are you using QueryParams instead?", this.querySpecDeserializer);
-		((DefaultQuerySpecDeserializer) this.querySpecDeserializer).setDefaultLimit(defaultPageLimit);
+		if (this.querySpecDeserializer instanceof DefaultQuerySpecDeserializer) {
+			((DefaultQuerySpecDeserializer) this.querySpecDeserializer).setDefaultLimit(defaultPageLimit);
```
remmeier: potentially it should rather throw an exception if it is not possible to set the page limit

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
```diff
@@ -431,23 +456,28 @@ public void setDefaultPageLimit(Long defaultPageLimit) {
 	 * NOTE: This using this feature requires a {@link QuerySpecDeserializer} and it does not work with the
 	 * deprecated {@link QueryParamsBuilder}.
 	 */
+	@Deprecated
```
remmeier: same as above, still in heavy use. Some kind of replacement would be needed.

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
```diff
@@ -66,7 +66,7 @@
 
 	private QueryParamsBuilder queryParamsBuilder;
 
-	private QuerySpecDeserializer querySpecDeserializer = new DefaultQuerySpecDeserializer();
+	private QuerySpecDeserializer querySpecDeserializer;
```
remmeier: set currently breaks with the the setter of default/max limit, where did you run into issues? should be ok to have the default one?

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
```diff
@@ -161,6 +163,26 @@ public void boot() {
 		bootDiscovery();
 	}
 
+	private void setupQuerySpecDeserializer() {
+		if (this.querySpecDeserializer == null) {
```
remmeier: maybe two ifs would be better, one to setup the default one. And the other to set all the properties regardless whether the default instance or a custom one is in use. If the non-custom one is in use and people do not want that. The application can simply ignore the properties.

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
```diff
@@ -431,23 +456,28 @@ public void setDefaultPageLimit(Long defaultPageLimit) {
...
+		}
 	}
 
 	/**
 	 * Sets the allow unknown attributes for API requests.
 	 *
 	 * NOTE: Recommend to follow JSON API standards, but this feature can be used for custom implementations.
 	 */
+	@Deprecated
 	public void setAllowUnknownAttributes() {
```
remmeier: same as above

crnk-meta/src/test/java/io/crnk/meta/integration/AbstractMetaJerseyTest.java
```diff
@@ -66,6 +71,7 @@ public TestApplication() {
 			objectMapper.enable(SerializationFeature.INDENT_OUTPUT);
 			feature.addModule(createModule());
 			boot = feature.getBoot();
+			boot.setQuerySpecDeserializer(new DefaultQuerySpecDeserializer());
```
remmeier: that should not be neccessary => backward compatibility

crnk-spring/src/main/java/io/crnk/spring/boot/v3/CrnkConfigV3.java
```diff
@@ -49,13 +57,22 @@ public CrnkBoot crnkBoot(SpringServiceDiscovery serviceDiscovery) {
 		CrnkBoot boot = new CrnkBoot();
 		boot.setObjectMapper(objectMapper);
 
+		/**
+		 * Try to get an existing `QuerySpecDeserializer` implementation and set it up.
+		 * Otherwise `DefaultQuerySpecDeserializer` instance used.
+		 */
+		try {
+			QuerySpecDeserializer querySpecDeserializer = this.applicationContext.getBean(QuerySpecDeserializer.class);
```
remmeier: this can be moved into CrnkBoot using ServiceDiscovery instead of applicationContext. This way it also works with CDI, Guava, etc.

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
vicmosin: That's actually was my main problem. I wanted to oeverride the `QuerySpecDeserializer ` impl and couldn't because of this explicit definition. And that's the main intention of this PR actually.

crnk-meta/src/test/java/io/crnk/meta/integration/AbstractMetaJerseyTest.java
vicmosin: This is necessary since I can only set default impl in boot method, but here it's not called yet

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
remmeier: with the updated spring config in this PR that should not be an issue anymore, even if there is still default instance that will then be overridden?

vicmosin: I don't understand your question.. currently `QuerySpecDeserializer` is defined explicitly so the only way to override it is to override the whole CrnkBoot class. There is a setter though but it's useless anyway because of `checkNotConfiguredYet()` check. What I wanted to implement is the dynamic instance creation (DI). This way anyone who wants his version of deserializer will just define a bean of appropriate type, default one will be used as a fallback mechanism.

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
vicmosin: Could you specify please? The idea that no need to configure Crnk to use default instance, it will be used implicitely if no custom bean found

crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
remmeier: I will take a closer look the next two, three days (today already occupied)

vicmosin:
@remmeier Any thoughts about the PR? Should I try to fix it or it's irrelevant from your side?

vicmosin:
@remmeier just quick reminder why PR was introduced. Main idea is to allow easy QuerySpecDeserializer injection. Currently it is hardcoded and you can't really inject your own instance since it will require to override the CrnkBoot class.

remmeier:
I merged https://github.com/crnk-project/crnk-framework/pull/201. I think that should contain the major issue (a comfortable way of injecting a custom serializer). What is now left to do is actually cleaning up the configuration/property api, which you here already deprecated.

remmeier:
also available on maven central http://mvnrepository.com/artifact/io.crnk/crnk-core

vicmosin:
@remmeier works as expected, thanks. You can then delete this PR

PR #167, opened 2017-11-24
Add possibility to serialize links as JSON objects

gianinbasler:
According to the JSON API specifications, links can either be serialized to simple JSON attributes, or JSON objects.
If Crnk Property "crnk.config.serialize.object.links" is set to true, links will be serialized to JSON objects.
Default behaviour remains unchanged, i.e. links are serialized to simple JSON attributes.


crnk-core/src/main/java/io/crnk/core/engine/internal/jackson/ObjectLinkErrorDataSerializer.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.IOException;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import io.crnk.core.engine.document.ErrorData;
+
+/**
+ * @author AdNovum Informatik AG
+ */
+public class ObjectLinkErrorDataSerializer extends ErrorDataSerializer {
```
remmeier: I noticed that at a number of locations with now have to implementations, one for primitive links and ones for object links. Potentially it would rather make sense to introduce an abstraction for writing links? then a single implementation at the various locations like error data and ObjectLinkResourceMapper would suffice.

remmeier:
crnk-client will need an update as well once crnk-core is finalized

remmeier:
thx!

PR #169, opened 2017-11-27
Handle 'Accept: application/json' messages as JSON-API requests by default

chklauser:
Adapt the JsonApiRequestProcessor and the CrnkServlet to accept HTTP
requests with 'Accept: application/json' (instead of
'application/vnd.api+json') too. This makes it easier to integrate with
tools that are not aware of JSON-API specifically.

To revert to the old behaviour, enable the configuration property
'crnk.config.resource.request.rejectPlainJson'.

Motivation: Yes, the JSON-API standard mandates that both client and server use the `application/vnd.api+json` Accept/Content-Type headers. But for GET requests, we think it is fine to send back 'application/vnd.api+json' even when the client has requested 'application/json'. This makes working with tools/frontend that are not aware of JSON API specifically easier (generic REST client libraries; REST testing tools, etc.) Sure, all of them can be configured to send `Accept: application/vnd.api+json`), but it needs to be done every single time with no real benefit.

Does not affect POST/PATCH requests. There, we want to be sure that the client is aware that it needs to format its payload as a proper JSON-API response.

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/boot/CrnkProperties.java
+++ b/crnk-core/src/main/java/io/crnk/core/boot/CrnkProperties.java
@@ -143,5 +143,19 @@ public class CrnkProperties {
 	 */
 	public static final String SERIALIZE_LINKS_AS_OBJECTS = "crnk.config.serialize.object.links";
 
-
+	/**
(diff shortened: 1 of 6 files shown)


remmeier:
thx!

PR #185, opened 2017-12-14
WIP: add throws declaration

### LINKED ISSUE #184: TransactionRunner should be able to throw Exceptions
The current `TransactionRunner` interface lacks of the ability to throw Exceptions. This breaks the functionality to handle exceptions via `ExceptionMapper`

In my specific task im trying to achieve the following result:
request: `delete /some/endpoint/123`
response: `403 data: {"errors":[{"id":"insufficient-permissions","title":"no permission","detail":"you dont have enough permissions","meta":{"module":"test","permission":"unlock"}}]}`

```java
public class InsufficientPermissionsException extends GlobalError {
    public InsufficientPermissionsException(PermissionInterface permission) {
        super(403, "insufficient-permissions", "no permission",
                String.format("you dont have enough permissions (%s.%s).",
                        permission.moduleName(), permission.permissionName()), createMeta(permission));
    }

    private static Map<String, Object> createMeta(PermissionInterface permission) {
        HashMap<String, Object> map = new HashMap<>();
        map.put("module", permission.moduleName());
        map.put("permission", permission.permissionName());
        return map;
    }
}
```

```java
public class LockEndpoint extends APIEndpointV2<Lock, Integer> {

    @Inject private LockTable lockTable;

    @Override
    public void delete(Integer id) {
        Lock dbLock = lockTable.findById(id);

        if (dbLock == null) {
            return;
        }

        if (!getUser().getId().equals(dbLock.getUserid())
                && !getUser().hasFunction(Permission.UNLOCK)) {
            throw new InsufficientPermissionsException(Permission.UNLOCK);
        }

        lockTable.delete(dbLock);

        if (!getUser().getId().equals(dbLock.getUserid())) {
            dbLock.getData().setLock(null);
            lockTable.createLock(dbLock.getData());
        }
    }
}
```

```java
public class ServerExceptionMapper implements ExceptionMapper<GlobalError> {

    public static final int HTTP_ERROR_CODE = 500;

    @Override
    public ErrorResponse toErrorResponse(GlobalError globalError) {
        return ErrorResponse.builder()
                .setStatus(globalError.getHttpStatus())
                .setSingleErrorData(globalError.getErrorData())
                .build();
    }

    @Override
    public GlobalError fromErrorResponse(ErrorResponse errorResponse) {
        return null;
    }

    @Override
    public boolean accepts(ErrorResponse errorResponse) {
        return false;
    }
}
```

under "normal" circumstances this would work pretty well. But if I'm using crnk-jpa it doesn't work anymore.

The problem is how the requests are handled inside of crnk. If you are registering jpa modules, every request seems to travel through every jpa module that is registered. the implementation of the jpa module requires you to implement the `TransactionRunner` interface. Every request seems to pass this method wehter the endpoint is mapped by crnk-jpa or not.

https://github.com/crnk-project/crnk-framework/blob/e7412436a34cda3200a25550528ad7fabf80ebbd/crnk-core/src/main/java/io/crnk/core/engine/internal/http/HttpRequestProcessorImpl.java#L93-L118

Thus the interface has no throws declaration and the `callable.call()` method has a throws declaration you have to try catch inside your implementation. this prevents any exception to "bubble up" and getting handled by the `ExceptionMapper`

Solution:
Add a `throws Exception` declaration to  `TransactionRunner`

mohlek:
fix #184

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/engine/transaction/TransactionRunner.java
+++ b/crnk-core/src/main/java/io/crnk/core/engine/transaction/TransactionRunner.java
@@ -4,5 +4,5 @@ import java.util.concurrent.Callable;
 
 public interface TransactionRunner {
 
-	<T> T doInTransaction(Callable<T> callable);
+	<T> T doInTransaction(Callable<T> callable) throws Exception;
 }
--- a/crnk-jpa/src/main/java/io/crnk/jpa/JpaModule.java
+++ b/crnk-jpa/src/main/java/io/crnk/jpa/JpaModule.java
@@ -370,13 +370,12 @@ public class JpaModule implements InitializingModule {
 
 			@Override
 			public Response filter(final DocumentFilterContext context, final DocumentFilterChain chain) {
-				return transactionRunner.doInTransaction(new Callable<Response>() {
-
-					@Override
-					public Response call() throws Exception {
-						return chain.doFilter(context);
-					}
-				});
+				try {
+					return transactionRunner.doInTransaction(() -> chain.doFilter(context));
+				} catch (Exception e) {
+					 logger.error("failed to execute operation", e);
+					 return null;
+				}
 			}
 		});
 	}
--- a/crnk-operations/src/main/java/io/crnk/operations/server/TransactionOperationFilter.java
+++ b/crnk-operations/src/main/java/io/crnk/operations/server/TransactionOperationFilter.java
@@ -4,11 +4,14 @@ import io.crnk.core.engine.internal.utils.PreconditionUtil;
 import io.crnk.core.engine.transaction.TransactionRunner;
 import io.crnk.core.module.discovery.ServiceDiscovery;
 import io.crnk.operations.OperationResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.List;
-import java.util.concurrent.Callable;
 
 public class TransactionOperationFilter implements OperationFilter {
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(TransactionOperationFilter.class);
 
 
 	@Override
@@ -17,11 +20,11 @@ public class TransactionOperationFilter implements OperationFilter {
 		List<TransactionRunner> transactionRunners = serviceDiscovery.getInstancesByType(TransactionRunner.class);
 		PreconditionUtil.assertEquals("expected single transaction runner", 1, transactionRunners.size());
 		TransactionRunner transactionRunner = transactionRunners.get(0);
-		return transactionRunner.doInTransaction(new Callable<List<OperationResponse>>() {
-			@Override
-			public List<OperationResponse> call() throws Exception {
-				return chain.doFilter(context);
-			}
-		});
+		try {
+			return transactionRunner.doInTransaction(() -> chain.doFilter(context));
+		} catch (Exception e) {
+			LOGGER.error("failed to execute operation", e);
+			return null;
+		}
 	}
 }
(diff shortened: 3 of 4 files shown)


mohlek:
Somehow this test fails for me

https://github.com/crnk-project/crnk-framework/blob/c7745863a980ae341a93ce3978d579dc0be5f289/crnk-jpa/src/test/java/io/crnk/jpa/integration/JpaQueryParamsEndToEndTest.java#L98-L126

```
io.crnk.jpa.integration.JpaQueryParamsEndToEndTest > testOptimisticLocking FAILED
    io.crnk.core.exception.ResourceNotFoundException
        at io.crnk.core.engine.internal.exception.CrnkExceptionMapper.fromErrorResponse(CrnkExceptionMapper.java:44)
        at io.crnk.core.engine.internal.exception.CrnkExceptionMapper.fromErrorResponse(CrnkExceptionMapper.java:16)
        at io.crnk.client.internal.ClientStubBase.handleError(ClientStubBase.java:130)
        at io.crnk.client.internal.ClientStubBase.execute(ClientStubBase.java:72)
        at io.crnk.client.internal.ResourceRepositoryStubImpl.executeUpdate(ResourceRepositoryStubImpl.java:52)
        at io.crnk.client.internal.ResourceRepositoryStubImpl.modify(ResourceRepositoryStubImpl.java:82)
        at io.crnk.client.internal.ResourceRepositoryStubImpl.save(ResourceRepositoryStubImpl.java:75)
        at io.crnk.jpa.integration.JpaQueryParamsEndToEndTest.testOptimisticLocking(JpaQueryParamsEndToEndTest.java:117)
```

I don't know how my changes result in this behavior.

Any ideas?

crnk-core/src/main/java/io/crnk/core/engine/transaction/TransactionRunner.java
```diff
@@ -4,5 +4,5 @@
 
 public interface TransactionRunner {
 
-	<T> T doInTransaction(Callable<T> callable);
+	<T> T doInTransaction(Callable<T> callable) throws Exception;
```
remmeier: not sure about this, can you not just catch and rethrow RuntimeExceptions and wrap the very rare checked exceptions into a runtime exception? because this change is slightly breaking.

crnk-jpa/src/main/java/io/crnk/jpa/JpaModule.java
```diff
@@ -370,13 +370,12 @@ protected void setupTransactionMgmt() {
...
-
-					@Override
-					public Response call() throws Exception {
-						return chain.doFilter(context);
-					}
-				});
+				try {
+					return transactionRunner.doInTransaction(() -> chain.doFilter(context));
+				} catch (Exception e) {
+					 logger.error("failed to execute operation", e);
```
remmeier: catching and not rethrowing the exception will cause problems

crnk-operations/src/main/java/io/crnk/operations/server/TransactionOperationFilter.java
```diff
@@ -4,24 +4,27 @@
...
-			@Override
-			public List<OperationResponse> call() throws Exception {
-				return chain.doFilter(context);
-			}
-		});
+		try {
+			return transactionRunner.doInTransaction(() -> chain.doFilter(context));
+		} catch (Exception e) {
+			LOGGER.error("failed to execute operation", e);
+			return null;
```
remmeier: same

crnk-core/src/main/java/io/crnk/core/engine/transaction/TransactionRunner.java
mohlek: yeah :-/ I think I'll give RuntimeException a try

mohlek:
wrapping with `RuntimeException` seems to work pretty well. But its still feels like a hack.
Maybe we can come up with a better solution. For now I can live with my RuntimeException wrapper and unwrapper :p

PR #192, opened 2018-01-11
enhance customizability

mohlek:
adds the ability to swap out classes more easily


mohlek:
@remmeier do you want me to write tests for these? Since the default implementations are tested anyways I don't see any need for this.

Issue #168, opened 2017-11-26
Spring Boot integration


remmeier:
currently crnk supports a basic spring integration with service discovery and jpa integration. 

goal would be to have full spring / spring boot integration:

- auto configuration for all modules
- spring cloud sleuth equivalent of crnk-monitor-brave
- allow json api exception handling in spring mvc
- metrics
- RestTemplate adapter for crnk-client
- Home module shows not only json api endpoint, but any kind of request mapping
- potentially expose actuators as json api endpoints

manosbatsis:
Both curious and interested, how would one go about implementing a more "native" spring integration, meaning generic spring-data based repo/service/controller implementations?

remmeier:
Spring Data is the one thing I personally rather avoided. I think in this area there Spring Data and JSON API are to different things with different goals/sementics addressing similar needs. Altough I guess it would be feasible to create a Spring Data JSON API similar to Spring Data Rest that provides at least some part of the JSON API feature set. But in terms of filtering, relationship handling, inclusions, etc.

manosbatsis:
I avoid Spring Data REST myself, but that is only because I find a 3 tier (repository, service, controller) approach more practical and extensible VS the springs single-tier approach. 

In short I would like to expose parts of crnk through such generic spring components, to provide a solid JSON-API base implementation but also allow better integration and extensibility within the spring ecosystem. I suppose I'll give it a try and come back with more specific questions.

remmeier:
closing this issue has quite a number of auto configurations are not available. In the future more specific tickets may provide further improvements.

PR #201, opened 2018-01-21
possibility to inject custom QuerySpecDeserializer #168

remmeier:


remmeier:
@vicmosin this would allow to inject a deserializer though the regular service discovery.

PR #206, opened 2018-01-30
chore: fix import & inline variables

mohlek:


remmeier:
somehow my PR scrambled up this one since this one is based on mine. The imports if IncludeLookupUtils have been the issue, right? I updated my Intellij settings, not sure why Intellij would think that using import something.* is a good idea...

mohlek:
was the same as in #228

PR #219, opened 2018-02-11
Issues/47

vicmosin:
Main changes:

- `PagingSpec` interface and its default implementation `OffsetLimitPagingSpec`: define particular paging strategy including paging data (i.e. limit/offset vars) and actual first/next/prev/last links generation

- `PagingSpecDeserializer` interface and its default implementation `OffsetLimitPagingSpecDeserializer`: provide abstraction for data deserialization for each `PagingSpec` strategy


remmeier:
looks pretty much perfect I would say :-) thx a lot. 

DefaultQuerySpecDeserializer.Parameter within the PagingSpec is maybe a bit ugly. But that is an issue of the current DefaultQuerySpecDeserializer. That one should have a minor cleanup.

there are two, three get/set offset/limit methods that may should be deprecated reather rather than removed directly. Altough not sure if people access DefaultQuerySpecDeserializer and QueryAdapter directly. Maybe not such much. So it would not be an issue.

remmeier:
one conceptual thing that we have to look out for: different repositories might follow different paging strategies. So ResourceInformation could carry a PagingBehavior. And the PagingBehavior provides a serializer, deserializer and maybe other things. Maybe the @JsonApiResource annotation could then have a new attribute pagingBehavior to set the paging behavior for that resource. Otherwise some global default with hold that may or may not be configurable.

vicmosin:
Yeah, that's actually good point. But I am affraid it will require to make deep changes in serialization/deserialization in general what I wanted to avoid..

remmeier:
the parameter holds a ResourceInformation, maybe it is already quite close to that.

remmeier:
as a warning: it is currently NOT possible to remove that crnk Optional. The reason for it has been the support of Java 7 till about two months ago. But now removing that optional is unfortunately a major break which would force the release of a Crnk 3.0

remmeier:
a would expect a Crnk 3.0 in about half a year.

vicmosin:
You gonna be kidding me..

remmeier:
it is a major break, so deprecation & new API is necessary as annoying as it is unfortunately.

vicmosin:
So I refactored a bit my initial PR. Had no time to get deeper into `PagingBehavior` concept. There are some errors as well in other modules, have no idea for now how to inject dynamic instance of `PagingSpecSerializer` :) Any feedback is welcome

vicmosin:
Mh.. I guess I will continue with it tomorrow morning as it's required for one of my projects.

vicmosin:
@remmeier 
Here is the first concept. Some tests are still failing, but I would like to hear your opinion on that. Example of usage you can see at `TaskWithPagingBehavior` class somewhere in tests. I wasn't sure which class should be responsible for filling out `ResourceInformation`'s PagingSpec serialization/deserialization and first used `DefaultInformationBuilder`. But then I realized that other modules such as client use `DefaultResourceInformationProvider` as a default mechanism and moved the logic there.

remmeier:
DefaultResourceInformationProvider  is the correct place. 

Maybe it would be worthwhile to merge serializer and deserializer. Something I also consider for the QuerySpec(De)Serializer.

ideally we rebase the actual commits without the Optional changes and merges once you think you are done. Then I can take a closer look & merge.

Issue #150, opened 2017-11-07
[crnk-jpa] filter by null value


mohlek:
What is the best practice to filter a field by `NULL` ?
In my tables there are plenty of fields with a `NULL` value. I'd like to request a resource with filter[lastEdited] = NULL` . Is there any way to do so?

mohlek:
@buschtoens

remmeier:
I guess that is more a general QuerySpec question that is currently unresolved. resp. the json api spec does not have recommendation on that matter. DefaultQuerySpecDeserializer.deserializeFilter would be the place. I suspect crnk-jpa will then properly pick the null up.

one possiblitiy would be to explicitly map the value "null" to null. There should be a possiblity to distinguish between null and empty string.

OskarD:
@remmeier any progress on this? Is there some way I can help?

remmeier:
nothing from my side so far. PRs are always welcomed. This one would be a fairly small one. Altough maybe we should set up something more generic? so allows allow other expressions next to "null", like "now " or "now + 1h" for dates and stuff. Is there a good existing expression language we could make use of maybe?

mohlek:
> so allows allow other expressions next to "null", like "now " or "now + 1h" for dates and stuff. Is there a good existing expression language we could make use of maybe?

I like the idea but not as part of the core. I think crnk should stick to the spec as close as possible.

OskarD:
As much as I wish that the spec covered this, I have to agree with @mohlek 

But I don't know how to do this myself, unfortunately, so I am hoping that someone else will ❤️

remmeier:
yes, we can also just get started with that null.

mohlek:
:heart_eyes:

PR #246, opened 2018-03-08
Support parameters that don't match target attribute type

weaselmetal:
behavior can be turned on via

    boot.getModuleRegistry().getTypeParser().setMismatchedTypeAcceptedAsString(true);

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,6 +56,7 @@ gradle.beforeProject { Project project ->
 		def ui = project.name == 'crnk-ui'
 		def testProject = project.name == 'crnk-test'
 		def legacyBraveProject = project.name == 'crnk-brave'
+		def bom = project.name == 'crnk-bom'
 		def examples = project.name.contains('example')
 
 		if (!docs) {
@@ -64,32 +65,34 @@ gradle.beforeProject { Project project ->
 			sourceCompatibility = 1.8
 			targetCompatibility = 1.8
 
-			dependencies {
-				testCompile group: 'junit', name: 'junit', version: '4.12'
-				testCompile group: 'org.mockito', name: 'mockito-core', version: '1.10.19'
-				testCompile group: 'org.assertj', name: 'assertj-core', version: '2.2.0'
-			}
-
-			test {
-				testLogging {
-					exceptionFormat = 'full'
-				}
-			}
-
-			apply plugin: 'io.spring.dependency-management'
-			dependencyManagement {
-				imports {
-					mavenBom 'org.springframework.boot:spring-boot-dependencies:1.5.8.RELEASE'
-					mavenBom 'io.projectreactor:reactor-bom:Bismuth-RELEASE'
-					mavenBom 'org.glassfish.jersey:jersey-bom:2.26'
-				}
+			if(!bom) {
 				dependencies {
-					dependency 'com.fasterxml.jackson.core:jackson-databind:2.9.4'
-					dependency 'com.fasterxml.jackson.core:jackson-annotations:2.9.4'
-					dependency 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.9.4'
-					dependency 'com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider:2.9.4'
-					dependency 'com.fasterxml.jackson.jaxrs:jackson-jaxrs-base:2.9.4'
-					dependency 'com.fasterxml.jackson.core:jackson-core:2.9.4'
+					testCompile group: 'junit', name: 'junit', version: '4.12'
+					testCompile group: 'org.mockito', name: 'mockito-core', version: '1.10.19'
+					testCompile group: 'org.assertj', name: 'assertj-core', version: '2.2.0'
+				}
+
+				test {
+					testLogging {
+						exceptionFormat = 'full'
+					}
+				}
+
+				apply plugin: 'io.spring.dependency-management'
+				dependencyManagement {
+					imports {
+						mavenBom 'org.springframework.boot:spring-boot-dependencies:1.5.8.RELEASE'
+						mavenBom 'io.projectreactor:reactor-bom:Bismuth-RELEASE'
+						mavenBom 'org.glassfish.jersey:jersey-bom:2.26'
+					}
+					dependencies {
+						dependency 'com.fasterxml.jackson.core:jackson-databind:2.9.4'
+						dependency 'com.fasterxml.jackson.core:jackson-annotations:2.9.4'
+						dependency 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.9.4'
+						dependency 'com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider:2.9.4'
+						dependency 'com.fasterxml.jackson.jaxrs:jackson-jaxrs-base:2.9.4'
+						dependency 'com.fasterxml.jackson.core:jackson-core:2.9.4'
+					}
 				}
 			}
 		}
@@ -229,8 +232,6 @@ gradle.beforeProject { Project project ->
 				tasks.publish.dependsOn(tasks.uploadArchives)
 			}
 			else {
-				apply plugin: 'maven-publish'
-
 				task sourcesJar(type: Jar) {
 					from sourceSets.main.allSource
 					classifier = 'sources'
--- a/crnk-bom/build.gradle
+++ b/crnk-bom/build.gradle
@@ -1,0 +1,38 @@
+description = "Crnk extensions BOM"
+
+def bomGenerator = {
+	def node = asNode()
+	def dependencyMgmtNode = node.appendNode('dependencyManagement')
+	def dependenciesNode = dependencyMgmtNode.appendNode('dependencies')
+
+	// add subprojects to the bom
+	project.rootProject.subprojects.each {
+		System.out.println(it.name)
+		if (it.name != 'crnk-bom') {
+			Node dep = dependenciesNode.appendNode('dependency')
+			dep.appendNode('groupId').value = it.group
+			dep.appendNode('artifactId').value = it.name
+			dep.appendNode('version').value = project.version
+		}
+	}
+}
+
+uploadArchives {
+	repositories {
+		mavenDeployer {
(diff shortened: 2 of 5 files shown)


crnk-core/src/main/java/io/crnk/core/queryspec/DefaultQuerySpecDeserializer.java
```diff
@@ -175,6 +175,11 @@ protected void deserializeFilter(QuerySpec querySpec, Parameter parameter) {
 		Class<?> attributeType = getAttributeType(querySpec, parameter.attributePath);
 		Set<Object> typedValues = new HashSet<>();
 		for (String stringValue : parameter.values) {
+			if (!typeParser.supports(attributeType) && typeParser.isMismatchedTypeAcceptedAsString()) {
```
mohlek: I would recommend to swap these two because of execution order.

crnk-core/src/main/java/io/crnk/legacy/internal/DefaultQuerySpecConverter.java
```diff
@@ -108,6 +108,11 @@ private void applyFilter(QuerySpec querySpec, String parameterName, Set<String>
 		Class<?> attributeType = PropertyUtils.getPropertyClass(querySpec.getResourceClass(), attributePath);
 		Set<Object> typedValues = new HashSet<>();
 		for (String stringValue : stringValues) {
+			if (!typeParser.supports(attributeType) && typeParser.isMismatchedTypeAcceptedAsString()) {
```
mohlek: The same applies to this one

crnk-core/src/main/java/io/crnk/core/queryspec/DefaultQuerySpecDeserializer.java
weaselmetal: can do, but the supports method is rather cheap too. from a semantic point of view, I found that its the more important check. but if you insist, I'm fine with swapping them

mohlek: @remmeier has to decide. I was just suggesting :)

weaselmetal: anyway, this is commit is more of a check if this is a route we can / should go down.

weaselmetal:
Not sure, does the coverage test fail because the number went down?

weaselmetal:
I'd like to get some feedback from you @remmeier

remmeier:
later today i have time. yes, if coverage goes down too much, it will prevent merging pr s

weaselmetal:
Can you please give me a hint how i could test the change? The methods that call parser.supports() is not covered by a test it seems. Doesn't make much sense to test supports() method and the new getter and setter.

remmeier:
on a second lookg, the supports() is actually to simplistic. Next to the registered parsers, it has also quite a number of defaults/fallbacks like parsing to Enums, looking for a static parse(String) method, etc. So this makes it rather hard to implement the supports with duplicating the work. One approach would be to introduce a second parse method that returns an Optional. Or all does fallback would have to be put as well into that parser map (once first looked up, the second time it rembers whether there is a parser for that or not).

weaselmetal:
Need to look at the code again regarding the Enum parsing.
I'm not a huge fan of the assumed parse method. I'd much rather communicate to register your own type parser, rather than look for a parse(String) method on the class. It is very non-obvious that crnk will look for such a method.
In any case, parsing will work just like it did when the new flag is not set.
I do agree that it ain't perfect yet, but it's a bit more flexible than before and solves a lot of issues for me.

weaselmetal:
How about a supports() method that does this:

    public boolean supports(Class<?> clazz)	{
		if (Enum.class.isAssignableFrom(clazz))
			return true;
		
		return parsers.containsKey(clazz);
	}

It would also be easy enough to implement a `isConstructableFromString(Class<?> clazz)` method that uses new methods `hasStringConstructor()` and `hasParseMethod()`.
Then those methods could be used in the `supports` method and in the actual `parse()` implementation.
But like I said before, it's not really transparent that crnk will look for certain constructors and methods.

If that is something that should be supported, then such types should probably implement predefined crnk interfaces. Eg.  something like

    interface StringParsable<T>
    {
        T parse(String value);
    }

@remmeier

remmeier:
not sure, but I think we should rather avoid that MismatchedTypeAcceptedAsString properly triggers a completely new parsing behavior. But agree that the current behavior is rather suboptimal and should be phased out in the long run. Has been this way for a long time since the old Katharsis days.

There is the SpringParser interface. StringParsable  is not doable since an interface cannot provide static methods.

remmeier:
maybe all the Java data/time data types can be added to DefaultStringParsers. And Enums. And then for others a warning will be printed that this is going to be removed in Crnk 3.0. Enums could be supported by this mechanism by also checking super types.

weaselmetal:
Sorry, I overlooked that `parse` should be a static method.

> all the Java data/time data types can be added to DefaultStringParsers

Not sure, define 'all the Java data/time data types'. This could become a long list. And, many applications (ours included), will use Joda for datetime stuff.
I tried to add an Enum parser to the DefaultStringParsers, but that doesn't work, as you need the final concrete Enum type, to be able to call `valueOf(String)`.

I believe it's fine (great even!) if crnk supports basic Java types (primitive types and Strings) and Enums out of the box, and nothing else.

remmeier:
we make heavy use of date/time in many applications (and more than likely most other users as well). I don't really want to tell to those applicaitons they now have to do this things manually. So slowly coming to conclusion to leave the logic as is. Otherwise at least Joda and stuff will cause some more work with detecting the presense of Joda and registering those types, etc. 

But what I would change is to make `Map<Class, StringParser> parsers` complete. Meaning that parse/supports check whether the requested type is already there. If not, the current logic is invoked and it will optionally provide parser. That might be a EnumStringParser(Class<Enum>...) or a ReflectionStringParser(Method methodToCall). That parser is then also put into the map. The second time parse/supports is invoked, it will then immediately know what to do. And parse/supports can share the codebase to get that parser. And your use case of having support() can easily be implemented. It would also help to speed-up the TypeParser a bit, altough it is currently not a bottleneck either.

weaselmetal:
Sounds all reasonable. My only question that's left:
Will I be able to retrieve unmappable parameters as String? That was the whole driver for this pull request. Our API depends on this ability. We need to retrieve the passed values for things like

- filter[person][company]=WhatEverIdentifiesACompanyButDontConstructACompanyInstance
- filter[person][hasSomeRelativeCalled]=Joe

Where `hasSomeRelativeCalled` is not a real physical attribute or method of a person.

This is another completely different topic, but we have API endpoints that aggregate data and return the results in jsonapi compatible format. But those endpoints depend on grouping parameters like
?group[salesStats]=month&group[salesStats]=user
... the way katharsis supported them.
Currently it's impossible to do that via QuerySpec. And since I know that I can get grip of the raw request along with all parameters, I gain way more flexibility by accessing the raw parameter map.

So, one thing that circles through my mind is: it would be great if crnk still provided me with all the passed parameters through QuerySpecs (if I turn that feature on) no matter if they map to a real attribute or if jsonapi felt like specifying them, OR, I'd like to be able to tell crnk "don't bother to parse any parameters, I'll do that myself from the raw request parameter map".
Letting the request parameters go through all the parsing and then do it all over again myself, leaving QuerySpecs unused, feels just wrong.

remmeier:
yes, that map in TypeParser would in the future hold a Map<Class, Optional<StringParser>> and supports(...) would accordingly check the optional. So if neither a parser has been specified nor any of the default parsers apply, the optional will be empty.

maybe let us not mix up to many things in one PR. group is not something specified by JSON API. But HttpRequestContextAware gives all the necessary flexiblity. If something in between would be needed. It would not be overly difficult to introduce a new repository interface and and query class. I work on the for the reactive things. A bit of modularity in this area may not hurt. Or maybe also more flexiblity in adapting the QuerySpec handling. But I'm not so sure about the grouping, that may should rather be a new resource type.

weaselmetal:
Oh, I just realized that [this commit](https://github.com/crnk-project/crnk-framework/commit/71a9f4c9655bf5fc77f5061b29ba0323e2d76bd2) also achieves somehow what I wanted to achieve.
By introducing allowUnknownParameters crnk doesn't choke on parameters any more that it doesn't understand. I tested our app against a build of current master without any adaptions and all my app's tests passed. Just for your info @remmeier.

weaselmetal:
just because I re-read our conversation:

> not sure, but I think we should rather avoid that MismatchedTypeAcceptedAsString properly triggers a completely new parsing behavior.

That part is not true. It's completely the same, unless someone decides that unmappable parameters should be accepted as String, but without being forced to ignore parse exceptions. And if someone changes that flag, then because of a certain intention. Don't know how you want to proceed @remmeier.

remmeier:
sorry, what I meant was that typeParser.supports(attributeType) currently does not honor all the fallback logic in place. So dates, etc. do not get parsed. The missing piece here is this I would say:

```
But what I would change is to make Map<Class, StringParser> parsers complete. Meaning that parse/supports check whether the requested type is already there. If not, the current logic is invoked and it will optionally provide parser. That might be a EnumStringParser(Class...) or a ReflectionStringParser(Method methodToCall). That parser is then also put into the map. The second time parse/supports is invoked, it will then immediately know what to do. And parse/supports can share the codebase to get that parser. And your use case of having support() can easily be implemented. It would also help to speed-up the TypeParser a bit, altough it is currently not a bottleneck either.
```

then two flags to allow unknown attributes (in place) and unknow types (this pr) would complement each other well. Maybe somethng like this could be added to @JsonApiResource annotation for even more fine-grained control.

If one would like to take it a bit further, I think the FilterOperator should also have a saying in what kind of value types are allowed. And some kind of expression language that can detect values like "null" (see https://github.com/crnk-project/crnk-framework/issues/150). Then I think we start being able to write application without having to ignore things in most to all cases.

weaselmetal:
Oops, seems I closed this PR by accident. Maybe we reopen it (which I can't do) such that the idea doesn't go away yet? @remmeier

Issue #256, opened 2018-03-22
DefaultResourceInformationProvider: accept method must also check against the JsonApiResource.pagingBehavior attribute


daniel-deluca:
When building the ResourceInformation, the ModuleRegistry iterates through all the builders it has and takes the first builders that accepts the resourceClass. 
The fact is that the first builder is always (seems to be) the ResourceInformationProvider (DefaultResourceInformationProvider) that contains, as pagingBehavior the OffsetLimitPagingBehavior. 
The DefaultResourceInformationProvider, in the public boolean accept(Class<?> resourceClass) method must check, not only, if the resourceClass contains the JsonApiResource annotation but also if the pagingBehavior value/class of that annotation is in the pagingBehaviors Map.

remmeier:
DefaultResourceInformationProvider must also be customizable, meaning having the possiblity to add further paging behaviors. That is already possible on the backend side through service discovery and CrnkBoot. But in CrnkClient offset/limit is still hard-coded. Maybe one should be able to register new paging behaviors through Module. That is something that works both on server and client side. Otherwise a CrnkClient.addPagingBehavior or something would be necessary.

daniel-deluca:
Through the usage of adding new module I can add the pagingBehavior but my issue description still is valid. I'm changing the code of io.crnk.core.engine.internal.information.resource.DefaultResourceInformationProvider#accept in order to solve this.

daniel-deluca:
Pull request done

remmeier:
yes, the fix would work. But there is a catch that the DefaultResourceInformationProvider is internal API. I think maybe DefaultResourceInformationProvider and PagingBehavior are two quite independent classes. For this reason I think we should try to get those additional pagingBehaviors into the existing DefaultResourceInformationProvider instead of having to create new ones. On the backend side that is already possible. On the frontend side not yet. It might also become rather hard to track down the isuse of a missing pagingBehavior registration if non of the information provides accepts a resource. I think it would be sufficient to throw an exception? Or do you see another issue?

daniel-deluca:
@remmeier : 
throwing an exception when non of the "Modules" are supporting the pagingBehavior specified class is OK for me as this means that there is a configuration/setup part missing.

With my fix in crnk-core, my client side code is just adding a new module with a specific ResourceInformationProvider that contains the pagingBehavior and adding the behavior class in the JsonAPIResource attribute.
```
[...]

PR #257, opened 2018-03-23
Issue 256

daniel-deluca:
The fix allow to accept the resource class only if the resource annotations contains JsonApiResource (like before) AND if the JsonApiResource.pagingBehavior value is part of the class pagingBehaviors.
If the test on pagingBehavior is not there, the OffsetLimitPagingBehavior will always be used no matter the pagingBehavior class set in the Resource's annotation

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/module/ModuleRegistry.java
+++ b/crnk-core/src/main/java/io/crnk/core/module/ModuleRegistry.java
@@ -39,6 +39,7 @@ import io.crnk.core.module.discovery.MultiResourceLookup;
 import io.crnk.core.module.discovery.ResourceLookup;
 import io.crnk.core.module.discovery.ServiceDiscovery;
 import io.crnk.core.module.internal.ResourceFilterDirectoryImpl;
+import io.crnk.core.queryspec.pagingspec.PagingBehavior;
 import io.crnk.core.repository.ResourceRepositoryV2;
 import io.crnk.core.repository.decorate.RelationshipRepositoryDecorator;
 import io.crnk.core.repository.decorate.RepositoryDecoratorFactory;
@@ -559,6 +560,12 @@ public class ModuleRegistry {
 
 		public ModuleContextImpl(Module module) {
 			this.module = module;
+		}
+
+		@Override
+		public void addPagingBehavior(PagingBehavior pagingBehavior) {
+			checkState(InitializedState.NOT_INITIALIZED, InitializedState.NOT_INITIALIZED);
+			aggregatedModule.addPagingBehavior(pagingBehavior);
 		}
 
 		@Override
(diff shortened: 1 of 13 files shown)


crnk-core/src/main/java/io/crnk/core/module/SimpleModule.java
```diff
@@ -1,14 +1,6 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
+import com.google.common.collect.ImmutableList;
```
remmeier: this will not work, we have in crnk-core only a compile-time dependency to org.reflections:reflections, which in turn has a dependency to guava. But reflections and with that also guava is not mandatory to use crnk (to keep it small/lightweight). So application would fail without it. But Collections.immutable... should do it.

crnk-core/src/main/java/io/crnk/core/module/SimpleModule.java
```diff
@@ -245,6 +244,20 @@ public void addJacksonModule(com.fasterxml.jackson.databind.Module module) {
 		return Collections.unmodifiableList(jacksonModules);
 	}
 
+	/**
+	 * Add the given {@link PagingBehavior} to the module
+	 * @param pagingBehavior the paging behavior
+	 */
+	public void addPagingBehavior(PagingBehavior pagingBehavior){
```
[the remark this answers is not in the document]
daniel-deluca: OK, got it, I've update it

crnk-core/src/main/java/io/crnk/core/module/SimpleModule.java
daniel-deluca: Fixed

daniel-deluca:
@remmeier : I've updated the code. Please check ... Thanks

crnk-core/src/main/java/io/crnk/core/module/ModuleRegistry.java
```diff
@@ -97,6 +87,8 @@
 
 	private ResourceFilterDirectory filterBehaviorProvider;
 
+	private List<PagingBehavior> pagingBehaviors = new ArrayList<>();
```
remmeier: potentially does is not necessary, since one can add new paging behaviors through a module

daniel-deluca:
@remmeier : Added a fix for #263

remmeier:
should we merge everything related to the paging first to separate the two things?

having the ResourcePath would be great as well to have. There is one limitation tough. ResourceInformation should not access it directly but rather pass it from DefaultResourceInformationProvider. Reason is that ResourceInformation is kept simple independent of any implementation. Not necessarly it must be backed by a resource class annotated with @JsonApiResource. crnk-jpa is one example, the dynamic repository mentioned in the docs another. 

At least the HomeModule will have to be touched as well (maybe there is some potential to reuse the url computation from resource registry). But I think ResourceRepository is otherwise fairly the only place I can currently think of.

daniel-deluca:
@remmeier I've reverted the changes in this PR related to issue #263, so you can merge the solution for issue #256

remmeier:
thx a lot. One minor issue left is I think CrnkClient.serviceDiscovery should be null by default, I will fix this later quickly up. Merged what we have already.

PR #268, opened 2018-04-05
crnk-client - make urlBuilder available

nicolabeghin:
dumb commit to make CrnkClient urlBuilder extendable

### DIFF
--- a/crnk-client/src/main/java/io/crnk/client/internal/ClientStubBase.java
+++ b/crnk-client/src/main/java/io/crnk/client/internal/ClientStubBase.java
@@ -141,4 +141,12 @@ public class ClientStubBase {
 	public enum ResponseType {
 		NONE, RESOURCE, RESOURCES
 	}
+	
+	public JsonApiUrlBuilder getUrlBuilder() {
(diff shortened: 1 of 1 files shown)


crnk-client/src/main/java/io/crnk/client/internal/ClientStubBase.java
```diff
@@ -141,4 +141,12 @@ protected RuntimeException handleError(HttpAdapterResponse response) throws IOEx
 	public enum ResponseType {
 		NONE, RESOURCE, RESOURCES
 	}
+	
+	public JsonApiUrlBuilder getUrlBuilder() {
```
remmeier: I think it would already be sufficient to be able to get/set the serializer, the JsonApiUrlBuilder is rather something internal I think people do not need to customize? currently it also  sits in the internal package. The getter/setters could then forward the call to JsonApiUrlBuilder.

nicolabeghin: you're right. deleting this PR, I'm gonna send another in a few minutes

Issue #284, opened 2018-04-19
Alternative pagination is hard to use - should become the global default


weaselmetal:
By now it should be possible to write your own implementation of `PagingBehavior`, which I did.
However, my support for `page[size]` and `page[number]` is not really picked up well.

My implementation is indeed used by crnk. It's a spring bean that crnk can get hold of via the nice `ServiceDiscovery` interface.
But, crnk stops creating pagination links in the response after registering this behavior. The reason is rather complicated.

The `JsonApiResource` annotation specifies that the default `pagingBehavior` is `OffsetLimitPagingBehavior.class`.
When the `ResourceInformation` for a resource is created, the following happens:
From the registered paging behaviors (only my custom one present), one is picked that matches the pagingBehavior of the annotated resources. [see this change](https://github.com/crnk-project/crnk-framework/commit/71a9f4c9655bf5fc77f5061b29ba0323e2d76bd2#diff-1dcde304f2fcd554078bc512586d5149R87)

Since it's my registered custom pagination versus the default OffsetLimitPagingation, on the `ResourceInformation` the `pagingBehavior` is set to null.
And finally [this check](https://github.com/crnk-project/crnk-framework/commit/71a9f4c9655bf5fc77f5061b29ba0323e2d76bd2#diff-742387973881b94147ab66bc183fdf3bR178) **stops crnk from writing any pagination links**.

I would have expected, that registering a custom pagination behavior ensures that this becomes the global default. That's not the case. I could go through all 80+ exposed resources and add the custom pagination behavior as desired behavior, but that ain't cool. In general, I'd want the same behavior for whole API. Maybe with the option to change it to something else where needed.

@vicmosin would it be possible to make a custom pagination behavior the new default behavior for the entire API?

remmeier:
too bad null is not a legal annotation value. I guess a DefaultPagingBehavior class is necessary that forwards to the global default.

remmeier:
already a number of annotation values go with a "DEFAULT" enum constant.

weaselmetal:
I think I have a solution. A noop-PagingBehavior as default, that will always get replaced by the first registered paging behavior should cut it. Gonna give it a try.

remmeier:
forwarding of DefaultPagingBehavior to the default implementation (available from the module registry) might be more transparent compared to replacing the implementation, depending on what is meant by replacing.

weaselmetal:
// cross check if desired resource paging behavior is a registered behavior, error out if not.
[...]

PR #288, opened 2018-04-20
Allow a custom paging behavior to become app default #284

weaselmetal:
Here's my PR

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/engine/internal/information/resource/DefaultResourceInformationProvider.java
+++ b/crnk-core/src/main/java/io/crnk/core/engine/internal/information/resource/DefaultResourceInformationProvider.java
@@ -1,13 +1,13 @@
 package io.crnk.core.engine.internal.information.resource;
 
 import java.lang.annotation.Annotation;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
 import io.crnk.core.engine.information.resource.ResourceField;
 import io.crnk.core.engine.information.resource.ResourceFieldInformationProvider;
 import io.crnk.core.engine.information.resource.ResourceInformation;
@@ -18,6 +18,7 @@ import io.crnk.core.engine.properties.PropertiesProvider;
 import io.crnk.core.exception.RepositoryAnnotationNotFoundException;
 import io.crnk.core.exception.ResourceIdNotFoundException;
 import io.crnk.core.queryspec.pagingspec.PagingBehavior;
+import io.crnk.core.queryspec.pagingspec.VoidPagingBehavior;
 import io.crnk.core.resource.annotations.JsonApiResource;
 import io.crnk.core.utils.Optional;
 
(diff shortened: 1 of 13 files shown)


crnk-core/src/main/java/io/crnk/core/engine/internal/information/resource/DefaultResourceInformationProvider.java
```diff
@@ -82,9 +83,21 @@ public ResourceInformation build(Class<?> resourceClass, boolean allowNonResourc
...
+							.filter(it -> pagingBehaviorType.isInstance(it))
+							.findFirst();
+
+			if (!optPagingBehavior.isPresent()) {
+				throw new IllegalStateException("paging behavior not found: " + pagingBehaviorType);
+			}
+
+		} else {
+			optPagingBehavior = pagingBehaviors.values().stream().findFirst();
 		}
```
weaselmetal: Those lines above would pretty much be the same, even if a delegating `DefaultPagingBehavior` was used (except that I'd have to instantiate a `DefaultPagingBehavior` that delegates to the actual implementation).
I decided against that, as it only makes it a little harder to figure out which behavior is used. @remmeier

weaselmetal:
please let me know what you think about that PR @remmeier

remmeier:
thx, what is the state regarding the forwarding vs null/replacing behavior? is forwarding still the way to go I think? Or is there an issue? If possible I would really like to simplify the crnk engine/setup part.

weaselmetal:
Take note of the review I started @remmeier. It's possible to delegate from a default behavior to a concrete behavior, but the code I commented on wouldn't change much. So I'd go for the replace idea, as you'd immediately see the concrete behavior.

Anyway, my last commit broke tests, and I wonder what the test expectation is.
Right now, in the test, a Custom paging behavior is registered against the framework.
But the test seems to expect, that a registered custom behavior is NOT used for classes, that don't specify any paging behavior. Rather, the default OffsetLimit paging behavior seems to be expected for such classes. But that is in conflict with what I was trying to achieve. So I guess way to go would be to adapt the tests?

remmeier:
I merged the other PR. Noop strategy is also fine for me as well. I think the current state of the PR already looks good.

Updating the tests is ok. Release notes will have to go a bit into detail. But more than likely it simplifies things for all people.

The PagingBehaviorDirectory would maybe still worthwhile to have to move that logic out of the DefaultInformationProvider (there is many things in there already). But not a priority for this PR.

weaselmetal:
Dang. service location through reflection on entire packages ain't cool.
VoidPagingBehavior, CustomOffsetLimitPagingBehavior (which is totally unrelated and irrelevant to the tests run afaik) and OffsetLimitPagingBehavior all are registered. By happenstance, the void behavior is then picked up (as it happens to be the first) and used a paging behavior on the resources.
Service location through reflection is going to lash back every single time an alternative (service) implementation is added and tests didn't expect that. Also, it (hopefully) never reflects how service location is set up in real projects. `ReflectionsServiceDiscovery` also hid the fact that setup of paging behavior through modules didn't work properly.
It may be convenient, but I thinks it's more a tool to shoot yourself in the foot.

remmeier:
I don't mind sending that ReflectionServiceDiscovery to hell for that test. Once upon a time it was the default Katharsis mechanisms. Nowadays it is still supported by not recommended (and deprecated at various places). But would not mind to drop it entirely because there are plenty of more reasonable/lightweight alternatives.

remmeier:
(but the dropping part would come with v3)

weaselmetal:
Would be great @remmeier if you could check if a build of that state works with your real world projects.

remmeier:
thx

PR #296, opened 2018-04-24
PR to fix #295

### LINKED ISSUE #295: JacksonModule registers incomplete pagingBehaviors
This is kind of a hen and egg problem.
Adding a module correctly depends on having pagingBehaviors registered already.
Otherwise `io.crnk.core.engine.internal.jackson.JacksonModule#setupModule` creates a new `DefaultResourceInformationProvider` instance with missing behaviors.
On the other hand, a module should be able to register new pagingBehavior. But it will lack any behavior that is added through modules that are added later on.
At the time when `io.crnk.core.engine.internal.information.resource.DefaultResourceInformationProvider#build(java.lang.Class<?>, boolean)` is called, the object may lack some registered paging behaviors (through modules) and fail.

I think the reason these tests worked:
https://travis-ci.org/crnk-project/crnk-framework/builds/370068790?utm_source=github_status&utm_medium=notification
was due to another bug / mishap: The serviceDiscovery is set to `ReflectionsServiceDiscovery(packageName)`, which picked up all the classes in the same package, including the `CustomOffsetLimitPagingBehavior`. Which I doubt was the intention. Maybe it was, anyway it hid the fact that it's not registered through the modules.

I haven't fully understood the entire bootstrapping process, but registering paging through modules behavior seems flawed. Is it not enough that any `PagingBehavior` is correctly registered through the `ServiceDiscovery`?

weaselmetal:


remmeier:
looks great, thx. ResourceInformationProviderModule  is a good way to go. That on can be made a bit more customizable at some point in the future.

PR #302, opened 2018-04-28
Return Bad Request exception when relationship resource type doesn't exist

philip-ince:
If you try and POST a resource with a relationship attached and the type doesn't exist, then you get a NPE.

I've changed the Invalid Relationship exception to a BadRequestException too because it's a user error


philip-ince:
#303

philip-ince:
Isn't 404 only appropriate if the server can't find what the URI refers to? In this case the server knows what to do with the URI but can't process the body. For that reason I'd say a 400/422 is more appropriate.

remmeier:
right, makes sense. I guess 400 is perfectly fine that. One small test case would be good to make sure things stay like this. Maybe somewhere in the area of RelationshipsResourcePostTest?

remmeier:
ping :-)

philip-ince:
ok @remmeier, i've adapted the changes according to @julillosamaral 's PR

PR #309, opened 2018-05-07
debug messages

mohlek:


mohlek:
Please be so kind and release a new patch level release.

remmeier:
will do a release, maybe collect a few more things the next one, two days

PR #314, opened 2018-05-09
Added self link to resource object

jharmelink:
Added self link to resource object according to the [JSON API spec](http://jsonapi.org/format/#document-top-level).


remmeier:
thx!

PR #339, opened 2018-06-13
fix ResourceNotFoundException on relationship beeing null

mohlek:


mohlek:
@remmeier does this look good to you?

mohlek:
any progress on this one?

crnk-core/src/main/java/io/crnk/core/boot/CrnkProperties.java
```diff
@@ -178,4 +178,7 @@ private CrnkProperties() {
 	 * @since 2.4
 	 */
 	public static final String REJECT_PLAIN_JSON = "crnk.config.resource.request.rejectPlainJson";
+
+
+	public static final String THROW_RESOURCE_NOT_FOUND_WHEN_RELATED_IDS_IS_NULL = "crnk.config.serialize.relation.throwResourceNotFound";
```
remmeier: would THROW_EXCEPTION_IF_RELATED_RESOURCE_IS_MISSING be a better name? because it is not really null, but rather an "error" in the related repository since the related ID is not null, but defined, but nevertheless it fails to lookup the related resource. Maybe together with a comment that this is rather undesired and a placeholder should be used instead.

remmeier:
you added the tests, thx. Looks good know.

PR #359, opened 2018-07-06
Allow setting resource path for JPA entity through @JpaResource 

kosmonaffft:
Allow setting resource path for JPA entity through @JpaResource annotation on JPA entity.
Old behavior of JPA module not changed.


remmeier:
looks good, thx! Could you add one small test? like in JpaResourceInformationProviderTest and JpaQuerySpecEndToEndTest by changing or adding such a new annotated entity?

At some point (not in scope of this PR) it would also be great to get rid of the @JpaResource annotation and just make allow the use of the regular @JsonApiResource annotation. In the early days that was not possible, but now should be.

remmeier:
thx!

PR #373, opened 2018-07-19
Add support fo @JsonApiResource in JPA entities...

kosmonaffft:


kosmonaffft:
Looks like already implemented in upstream...

Issue #378, opened 2018-08-01
Attribute treated as a JsonApiRelationId even though not annotated


duncanportelli:
I have a Person resource which has a to-one relationship to the Address resource. In the Persons resource, apart from the relationship, I have a field `addressId`:
```
public class Person {

    private Integer addressId;
	
    @JsonApiRelation(lookUp = LookupIncludeBehavior.AUTOMATICALLY_ALWAYS, repositoryBehavior = RelationshipRepositoryBehavior.CUSTOM)
    private Address address;

}
```
Through multiple debug runs, it appears that even though the annotation `@JsonApiRelationId` is not present on the `addressId` attribute, Crnk still seems to treat it as a `JsonApiRelationId` due to its name (**relationship name + id suffix**), which results in unwanted as well as unexpected behaviour. This assumption is made in the [`ResourceInformationProviderBase`](https://github.com/crnk-project/crnk-framework/blob/master/crnk-core/src/main/java/io/crnk/core/engine/internal/information/resource/ResourceInformationProviderBase.java#L123) class.

This is causing the `ResourceRepository` to be used to find the addresses instead of the `RelationshipRepository`, something which I do not want due to certain logic. Is this standard functionality or is this an issue please? In my opinion, a field should only be considered as a JsonApiRelationId only if it is annotated and not just by its name.

Is there any way I can avoid this scenario without changing my variable name please?

remmeier:
Yes that is a mistake. It must check for the presence of @JsonApiRelationId

Koshux:
+1 - Thanks for your reply. When do you reckon a fix would be released for this?

remmeier:
If you have time for a PR to change that line I can merge it and it will automatically be released to the latest repository. This week I m off and don't have my notebook with me.

duncanportelli:
Thanks for your prompt reply @remmeier. Created a PR for this (https://github.com/crnk-project/crnk-framework/pull/379). Apologies if something is wrong (this is my first time :) )

PR #379, opened 2018-08-03
Attribute treated as a JsonApiRelationId even though not annotated

duncanportelli:
Fixed an issue by checking that a field has the JsonApiRelationId annotation before setting it as an id to a relationship field.

Refer to https://github.com/crnk-project/crnk-framework/issues/378 for more details.


remmeier:
thx

PR #387, opened 2018-08-10
Introduced patch strategy

### LINKED ISSUE #386: Allow define depth while merging resources during PATCH
As soon as resource has an object of type `JsonNode` (from jackson) merge operation iterates over the attributes . As a result, all the keys inside that custom field are also compared meaning if I send 
```
{ 
   "id": "1",
   "attributes": {
     "json": "{\"key\": \"value\"}"
   }
}
```
but original resource has json value of `{"key2": "value2"}` the resulted value which will go to storage will be
```
{ 
   "id": "1",
   "attributes": {
     "json": "{\"key\": \"value\", \"key2\": \"value2\"}"
   }
}
```
which it makes no sense as I want to override the existing value. Speicification also says nothing about the depth of the merging. Can we fix that?

vicmosin:
Resolves #386


vicmosin:
@remmeier not sure what happend there

remmeier:
looks like infrastructure problems somewhere with travis, not sure.

remmeier:
now things look well again. But there are compile errors.

remmeier:
that is QueryDSL which is generated by an annotation processor. depending on your ide, you may need to run the gradle build to have those classes generated.

vicmosin:
Ok, thanks. Fixed CI

remmeier:
looks perfect. There is a minor conflict, then we sequash and merge. if you like you can also add something to /crnk-documentation/src/docs/releases/latest.

remmeier:
thx

Issue #238, opened 2018-02-22
Support versioning trough content type


remmeier:
it is quite common to do some kind of versioning for REST api. Typically by introducing /api/v1/..., /api/v2/... directories. For resource-oriented API this is usually a bad match as linking must become versioned to. Alternative strategy is to:

1. incrementally add new fields and for clients to ignore things they do not know
2. create new resources with new name in case of a major break
3. if it really has to be: add version to content type and deliver proper version but maintain url

goal would be to support 3. 

Possiblities:
1. repositories would carry a version range (may or may not be open on one or both ends) through an annotation. ResourceRegistryPart is wrapped by a VersionedResourceRegistryPart that only returns repositories that match the request version range (trasnmitted trough content type).
2. ...?  

input very welcomed...

weaselmetal:
I think that reasoning which approach to pursuit should check against popular client JS frameworks such as angular, ember, etc - determining which approach is best supported. But I don't know any JS framework well enough.

weaselmetal:
Kinda old, but hopefully not invalid by now: http://andycrum.com/2014/06/02/getting-started-with-ember-data/
Look for: An aside: specifying a different host and namespace

corrspt:
`Ember Data` (Ember's Data Store) is currently at `3.0` (at the time of the article is was pre 1.0), but the content of that article is mostly accurate (somethings have changed, of course).

`Ember Data` (and Ember) by default use JSONAPI, through an adapter (named `JSONAPIAdapter`). Adapters allow you to customize the host/namespace and also headers. Here's a link to the current [Ember Data Documentation (3.0 at the time of writing)](https://guides.emberjs.com/v3.0.0/models/customizing-adapters/)

In Ember Data you define an application level adapter (by default `JSONAPI Adapter`) and can "override" the default adapter per-model, if you need. I'm thinking of doing this as way to progressively migrate away from Katharsis.

For reference customising the `path` so that requests are directed to `https://api.exameple.com/api/1`

```js
import DS from 'ember-data';

export default DS.JSONAPIAdapter.extend({
  namespace: 'api/1'
  host: 'https://api.example.com'
});
```

Or customizing headers:

```js
import DS from 'ember-data';

export default DS.JSONAPIAdapter.extend({
  headers: {
    'API_KEY': 'secret key',
    'ANOTHER_HEADER': 'Some header value'
  }
});
```

michalgce:
@remmeier did you handle versioning via content type ?

[...]

PR #393, opened 2018-08-23
WIP: [238] Support versioning trough content type

michalgce:
Looks like it works but for sure code needs improvements, would be cool if you could place some feedback.

Knows missing things:
- [ ] Tests not passing for external modules
- [ ] Validation of adding versioned resources (overlaps)

Related to #238

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
+++ b/crnk-core/src/main/java/io/crnk/core/boot/CrnkBoot.java
@@ -42,11 +42,7 @@ import io.crnk.core.engine.internal.utils.PreconditionUtil;
 import io.crnk.core.engine.properties.NullPropertiesProvider;
 import io.crnk.core.engine.properties.PropertiesProvider;
 import io.crnk.core.engine.query.QueryAdapterBuilder;
-import io.crnk.core.engine.registry.DefaultResourceRegistryPart;
-import io.crnk.core.engine.registry.HierarchicalResourceRegistryPart;
-import io.crnk.core.engine.registry.RegistryEntry;
-import io.crnk.core.engine.registry.ResourceRegistry;
-import io.crnk.core.engine.registry.ResourceRegistryPart;
+import io.crnk.core.engine.registry.*;
 import io.crnk.core.engine.result.ResultFactory;
 import io.crnk.core.engine.url.ConstantServiceUrlProvider;
 import io.crnk.core.engine.url.ServiceUrlProvider;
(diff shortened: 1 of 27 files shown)


remmeier:
A great start!. A first iteration:

- the Version annotation could be moved to the other annotations like JsonApiResource and share the JsonApi prefix.
- Currently undecided whether not to move the versioning information from ResourceRepositoryInformation to ResourceInformation. The current one is more flexible (different repository implementations but with the same resource structure are possible), while for ResourceInformation the resource itself must change to create a new version since it would be tied to the resource. The later would allow CrnkClient to work with the version annotation  and send the proper version headers. Have to think about that.
- The difficult part: Those ResourceRegistry.getEntry(...)/hasEntry(...) methods are used everywhere. So any change here has a gigantic impact. Maybe we should introduce a new method "ResourceRegistry getXy(QueryContext)" that returns something similar to ResourceRegistry but scoped to the current request and then can be used just like the ResourceRegistry. For this to work, QueryContext could be complemented by a new attribute "version" based on the transmitted header.  That new object no longer need any versioning information, since it is request-specific.  Most methods of the ResourceRegistry could then be phased out/deprecated or rather for the time being delegate to a current properly scoped object (there is a thread local to get to QueryContext, but works only for the sync crnk version, not async). For a start this delegation would allow to have full versioning support in sync repositories without breaking/changing any API.  The desigsn would also allow more flexiblity in the future about those matters since it is not tied to versioning and can be used for any kind of request-specific behavior. Already now it would also easy url access since all those methods already carry a QueryContext parameter that could then be eliminated on the scoped object.

makepanic:
Is there a specific reason why it was chosen to version the api by modifying/parsing the request `content-type` header?

There was a [long thread](https://github.com/json-api/json-api/issues/406) in the json-api issues around versioning and many different reasons pro/contra various approaches.

There are also approaches, [like stripe](https://stripe.com/blog/api-versioning), which completely hides the API version in some cases:

* A Stripe-Version header if one was supplied.
* The version of an authorized OAuth application if the request is made on the user’s behalf.
* The user’s pinned version, which is set on their very first request to Stripe.

I'm not that fluent with the crnk code but here are various questions I got when looking at the PR:

* is this pluggable so one could potentially reconfigure it to use a different pattern/approach?
* it looks like the header is only handled for usage in `Content-Type`, does this means this change won't effect api consumers as they would be using the `Accept` header?

I could also be completely wrong in how I understood the PR, so please excuse anything that's factual wrong 😅

remmeier:
in general the PR is the early stages, not sure how much time @michalgce has. Major challenge is more in the backend implementation. How the "current" version is read from the request headers in that regard is a minor issue and could very well be made puggable. For the request header, it could/should go into the accept header, but with other possiblities pluggable I guess.

Major issue that has to be addressed in the PR is how to make that version transparently available through all places in the crnk enginem, in particular the resource registry as the resource registry now depends on the requested version.

makepanic:
Thanks for the detailed explanation. You can ignore my previous comment then.

remmeier:
closed due to inactivity, feel free to reopen

PR #414, opened 2018-10-17
support for renamed attributes in error-response

### LINKED ISSUE #406: Renamed attributes are not reflected validation result's error object
Context: Spring MVC based project

```
class Resource {
    @NotBlank
    @JsonProperty("display-name")
    private String displayName;
}
```

When a client attempts to create this resource the server responds (422) with the following error object:

```
 {
  "errors" : [ {
    "status" : "422",
    "code" : "javax.validation.constraints.NotBlank",
    "detail" : "must not be blank",
    "source" : {
      "pointer" : "/data/attributes/displayName"
    },
   ...
  } ]
}
```

As you can see the JSON-pointer contains pojo's field-name instead of the "public" name

systemfreund:
Resolves #406 

In order to fix `ValidationMetaProviderTest` I had to modify `ResourceMetaPartition` and `MetaAttribute`. Not sure if that's the right approach, though.


remmeier:
thx!

Issue #396, opened 2018-08-29
Using different paging behaviors with included parameters gives errors


vicmosin:
`RepositoryAdapterUtils` fails with 

```
io.crnk.core.queryspec.pagingspec.OffsetLimitPagingSpec cannot be cast to io.crnk.core.queryspec.pagingspec.NumberSizePagingSpec
```

while 

```
if (pagingBehavior != null && pagingBehavior.isRequired(requestSpec.getQueryAdapter().getPagingSpec())) {
```

Reason is the following - `pagingBehavior` points to the current PagingBehavior object but `requestSpec.getQueryAdapter().getPagingSpec()` leads to the initial one which was provided with request. 

As a result, `pagingBehavior.isRequired(requestSpec.getQueryAdapter().getPagingSpec())` can't implicitely convert `PagingBehaviorA` to `isRequired(PagingBehaviorB)`

Simple solution could be adding additional check i.e. 

```
if (pagingBehavior != null && pagingBehavior.createEmptyPagingSpec().getClass().isInstance(requestSpec.getQueryAdapter().getPagingSpec()) && pagingBehavior.isRequired(requestSpec.getQueryAdapter().getPagingSpec())) {
```

to make sure both vars points to the same type.. But looks kinda odd..

@remmeier Any ideas?

remmeier:
sorry, I missed this ticket😔. i guess QueryAdapter.getPagingSpec does not make much sense in the first place. it should take a ResoureInformation as parameter and get it from that QuerySpec (inclusions have separate QuerySpec).

vicmosin:
@remmeier I am still kinda confused regarding this issue.. The main cause of the behaviour is `RepositoryAdapterUtils#l52` (if statement I mentioned on original issue) so it can be easily fixed with additional condition. But what I don't understand is whether it's a bug or wrong configuration from my side.

Issue is that `requestSpec.getQueryAdapter().getPagingSpec()` in check above is always instance of `OffsetLimitPagingSpec` because it is [set explicitely ](https://github.com/crnk-project/crnk-framework/blob/master/crnk-core/src/main/java/io/crnk/core/queryspec/QuerySpec.java#L56)at `QuerySpec` public constructor no matter which paging-spec is defined by resource itself. Is this correct?

remmeier:
yes it is definately a bug.

remmeier:
it is available on the latest repository, an official one next week again. Waiting for one other PR and potentially one, two things on my own.

[...]

PR #431, opened 2018-11-27
Potential fix of #396

vicmosin:


remmeier:
thx! do you have by chance a complete stack trace? But I think it is exactly this.

Could you change one thing? Avoiding the new constructor in favor of using the setter? QuerySpec is quite public and I would like to keep it as small as possible. If you do it in the ForwardingStrategyContext, you have access to ResourceInformation, RegistryEntry and ResourceRegistry. RegistryEntry holds the PagingBehavior and allows to create an empty one.

Issue #430, opened 2018-11-27
MetaLookup of generic members


lorenzleutgeb:
Hi!

Consider the following case of an entity that inherits from a superclass with generics:

```java
@MappedSuperclass
class Superclass<T> {
  @Column
  T foo;
  ...
  // Imagine some getters, setters and an empty ctor.
}

@Entity
class MyEntity extends Superclass<String> {
  ...
  // Imagine some @Id, other getters, setters and an empty ctor.
}
```

The discovery of `Superclass#foo` will not work: It will produce an `IllegalStateException` thrown by `io.crnk.core.engine.internal.utils.ClassUtils#getRawType` https://github.com/crnk-project/crnk-framework/blob/aa6497460633e1c0bcc07374b7d98b2d74aa0a97/crnk-core/src/main/java/io/crnk/core/engine/internal/utils/ClassUtils.java#L372-L380 since the type it receives is a [`sun.reflect.generics.reflectiveObjects.TypeVariableImpl`](https://hg.openjdk.java.net/jdk-updates/jdk11u/file/16d5ec7dbbb4/src/java.base/share/classes/sun/reflect/generics/reflectiveObjects/TypeVariableImpl.java).

### Case 1: This behaviour is intended and you do not want to change the situation anytime soon:

crnk's default implementation of `MetaProvider` does not handle this case on purpose. If so, then my request is to add the case of type variables to `getRawType` and throw a more specific exception with a message saying that generics are not supported. In addition, the documentation should mentation this.

### Case 2: You want to allow this case:

Otherwise, you would have to improve your logic and do some simple type inference: When you handle `Superclass#foo` the context of `MyEntity` and its concretization for `T` is of relevance. You will never be able to get the actual type of `foo` by looking at its decalaration locally. However it looks like this is exactly what is happening in your current implementation: In this case the `TypeVariableImpl` would just carry its name `"T"` and its bounds (which are none in my example).
Also, I am wondering why the superclass is even inspected so closely, since it is not an actual entity but only the superclass. My guess is that you get to the declaration in the superclass since it is also a declaration in the entity by inheritance. But I did not investigate so far into your code.

remmeier:
[...]

PR #432, opened 2018-11-27
Resolve concrete types of type variables

lorenzleutgeb:
This is an attempt to advance #430 by resolving the concrete types of type variables at runtime.

I got my `JpaModule` to initialize, however requests to the meta module still fail with the same exception as described in #430.


crnk-activiti/src/main/java/io/crnk/activiti/mapper/ActivitiResourceMapper.java
```diff
@@ -177,7 +177,7 @@ private void copyDynamicField(Object resource, BeanAttributeInformation attribut
 				if (toResource) {
 					Object value = variables.get(key);
 					if (value != null) {
-						Object mappedValue = mapValue(value, attribute.getImplementationClass());
```
remmeier: Unfortunately things cannot be renamed for the sake of backward compatiblity. The exception is everything within a "internal" package.

crnk-core/src/main/java/io/crnk/core/engine/information/bean/BeanAttributeInformation.java
```diff
@@ -160,10 +148,56 @@ public String getJsonName() {
...
+			return getter;
+		}
+		if (setter != null) {
+			return setter;
+		}
+		throw new IllegalStateException("Required: field, getter, or setter!");
+	}
+
+	public Class<?> resolveRawClass() {
+		return TypeResolver.resolveRawClass(this.getGenericType(), this.getBeanInformation().getImplementationClass());
```
remmeier: could you cache this? BeanInformation must be efficient to look-up for performance reasons.

crnk-core/src/main/java/io/crnk/core/engine/internal/utils/ClassUtils.java
```diff
@@ -395,4 +427,27 @@ public static Type getElementType(Type genericType) {
 		}
 		return genericType;
 	}
+
+	/**
+	 * Get the underlying class for a type, or null if the type is a variable type.
+	 * @param type the type
+	 * @return the underlying class
+	 */
+	private static Class<?> getRawTypeNullable(Type type) {
```
remmeier: you could make use of the Optional to make it explicit that there might be no raw type. Or should it be Object.class?

crnk-core/src/main/java/io/crnk/core/engine/internal/utils/PropertyUtils.java
```diff
@@ -76,6 +78,10 @@ public static Type getPropertyType(Class<?> beanClass, String field) {
 		return INSTANCE.findPropertyType(beanClass, field);
 	}
 
+	public static Class<?> getRawPropertyType(Class<?> beanClass, String field) {
```
remmeier: could you rely here on BeanInformation.of(...) rather than introducing more methods to this class? Then we could maybe sooner rather than later get rid of it... :-)

crnk-meta/src/main/java/io/crnk/meta/internal/typed/MetaDataObjectProviderBase.java
```diff
@@ -31,7 +32,9 @@ protected void createAttributes(T meta) {
 		for (String name : beanInformation.getAttributeNames()) {
 			BeanAttributeInformation attrInformation = beanInformation.getAttribute(name);
 			if (attrInformation.getGetter() != null && !isIgnored(attrInformation)) {
-				if (attrInformation.getGetter().getDeclaringClass() != implClass) {
+				boolean getterDeclaredInImplClass = attrInformation.getGetter().getDeclaringClass() == implClass;
+				boolean generic = attrInformation.getField().getGenericType() instanceof TypeVariable;
+				if (!getterDeclaredInImplClass && !generic) {
```
remmeier: I'm a bit undecided about this one. The direction is good. But not sure, I think we can rather not make a new "declarted" MetaAttribute in the subtype. Somehow it should rather be reflected in the "attributes" rather than "declaredAttributes" list.

crnk-core/src/main/java/io/crnk/core/engine/internal/utils/PropertyUtils.java
lorenzleutgeb: I do not understand what you mean by `BeanInformation.of(...)`. I cannot find such a method.

crnk-core/src/main/java/io/crnk/core/engine/internal/utils/ClassUtils.java
lorenzleutgeb: I wrote this because the `getRawType` throws on `TypeVariable`s, which I think is not sensible. And it just blew up everything. The idea was that a type variable itself just does not have a raw type. With this semantics, I would prefer `Optional.empty()`, i.e. no raw type.

However, there might be a way out: If we would resolve the concrete type for that type variable, then we could very well get its raw type, and I think that would make sense. Maybe I'll do that afterwards.

crnk-activiti/src/main/java/io/crnk/activiti/mapper/ActivitiResourceMapper.java
remmeier: yes, just public API, thx!

crnk-core/src/main/java/io/crnk/core/engine/internal/utils/ClassUtils.java
remmeier: feel free to deprecate the current method and introduce a new one. Or even better rely on the BeanInformation where possible.

crnk-core/src/main/java/io/crnk/core/engine/internal/utils/PropertyUtils.java
remmeier: ah, sorry BeanInformation.get(...)

remmeier:
there are some build issues. On master I downgraded to Gradle 4.x again. But there are some further memory issues that have to be further looked at. Seems like it reached a limit.

remmeier:
should I have another look at everything regarding my last comments? I plan to have a release together in a few days (2.11). So we can integrate this as well.

remmeier:
FYI: I will publish a new release this weekend if you like we can look at finishing this PR before

lorenzleutgeb:
Yes, I have some time today and tomorrow to work in any comments.

Obviously the weak part of this PR are the changes in `ClassUtils` which you will object to since you want to get rid of it. The functionality implemented there would IMO fit nicely into `net.jodah:typetools`. My idea is to merge here nevertheless and then upstream, since I do not think that getting `typetools` to merge and release these changes within the next two days is feasible.

lorenzleutgeb:
@remmeier FYI, jhalterman/typetools#46 is progressing, which will eliminate all the stuff I added to `ClassUtils`.

lorenzleutgeb:
@remmeier The PR already landed, `typetools:0.6.0` was released, and I updated this PR accordingly. Even without hurrying :)

remmeier:
thx, looks great.  Maybe two further checks would be good to have to verify that everything implemented works as desired. I thinkg this to checks could be added to testGenericsOnParentMappedSuperClass:

- that meta.getAttributes returns no duplicate attributes (not sure, may or may not be an issue currently) with both the parent and base types
- that the type parameters of the genericListSuper and genericList is properly resolved for the subtype.

remmeier:
one potential other issue is that currently it is verified in the context of JPA. Maybe it could make sense to add a test to BeanInformationTest and DefaultResourceInformationProviderTest to verify that it works in a non-JPA context.

If like you can also add a short snipped to DefaultResourceInformationProviderTest and/or the docs.

remmeier:
@lorenzleutgeb I rebased the PR, added a further test case and fixed an issue with duplicate MetaAttribute in getAttributes in case of generic super type attributes. release to follow shortly

Issue #447, opened 2018-12-27
Serialized comma-separated filter value can't be properly deserialized


vicmosin:
While building links, `JsonApiUrlBuilder` component encodes all serialized values so any comma-separated value becomes `x%2Cy` (which is `x,y` originally). As soon as one tries to follow that link, `DefaultQuerySpecUrlMapper` can't properly decode filter query spec as it expects value to be already decoded but the value isn't yet.

remmeier:
DefaultQuerySpecUrlMapper should not directly have to consider url/parameter decoding. That is the task of the underlying HTTP layer. I quickly started the Spring Boot application and requested with `http://127.0.0.1:8080/api/projects?page[limit]=10&filter[projects][id][EQ]=121%2C122` and `http://127.0.0.1:8080/api/projects?page[limit]=10&filter[projects][id][EQ]=121,122`. Both worked fine (with the later not being properly encoded, but the browser taking care of it similar to the whitespace charater). Subsequent links are then properly encoded with `%2C` and using those links works fine again. If you encounter an issue in this area, then it seems related to the HTTP implementation. Do you have a Spring Boot setup or someting else?

vicmosin:
Yes, spring boot.. and I get that link as prev/next link serialized so I can't follow it because it can't be deserialized with proper filter value

remmeier:
ok, once the above url with `%2C` hits `DefaultQuerySpecUrlMapper.deserialize`, `parameterMap` should hold again the seperator as `,` in its decoded form. I assume that is not the case. Those parameters come from `HttpRequestContextBase.getRequestParameters`. In the Spring Boot case by default this is provided by `ServletRequestContext` where the parameters are fetched in `getParameters`. Do you see where it goes wrong in your case?

vicmosin:
> ok, once the above url with %2C hits DefaultQuerySpecUrlMapper.deserialize, parameterMap should hold again the seperator as , in its decoded form

yes, it should but I get a single value as `x%2Cy` .

The only thing I have different is custom pagination but there I use internal url-builder

remmeier:
can you check the ServletRequestContext, that `getParameters` makes use of proper encodings? it probably is in this area. Not sure what triggers it to not working. Servlet Container is Tomcat or something else?

vicmosin:
So you are saying that `ServletRequestContext.getParameters` should properly convert `x%2Cy` to `x,y`?

remmeier:
[...]

PR #448, opened 2018-12-27
Potential fix of #447

vicmosin:


remmeier:
how do you mean parameters to be encoded in the QuerySpec? FilterPec.value holds by default regular Java types without any http/json specific encoding. multiple values are then represented as collection.

vicmosin:
@remmeier According to #447 ,`DefaultQuerySpecUrlMapper` doesn't take into consideration url-encoded values.

PR #449, opened 2018-12-27
replace crnk Optional with Java 8 counter-part

### LINKED ISSUE #225: replace crnk Optional with Java 8 counter-part
since Java 8 has become the minimum requirement, we can move from the Crnk Optional class to the default Java 8 one.

This means:
- Modify all internal uses accordingly
- Deprecate crnk optional
- Deprecate all uses and provide a Java 8 replacement
- Drop deprecated code with Crnk 3 (summer/fall to maintain a yearly major release cycle)

@vicmosin  FYI

systemfreund:
Resolves #225


remmeier:
thx

PR #452, opened 2019-01-07
Crnk spring actuator metrics support

### LINKED ISSUE #451: Prometheus support
i.e. metrics. Any work in that direction?

vicmosin:
Added custom tags-provider and implementation auto-configuration to define it, which properly resolves uri for any crnk resource

Partially resolves #451

### DIFF
--- a/crnk-setup/crnk-setup-spring-boot2/src/test/java/io/crnk/spring/metrics/CrnkWebMvcTagsProviderTest.java
+++ b/crnk-setup/crnk-setup-spring-boot2/src/test/java/io/crnk/spring/metrics/CrnkWebMvcTagsProviderTest.java
@@ -1,0 +1,78 @@
+package io.crnk.spring.metrics;
+
+import com.google.common.collect.Iterators;
+import io.crnk.core.engine.registry.RegistryEntry;
+import io.crnk.core.engine.registry.ResourceRegistry;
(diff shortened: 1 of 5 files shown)


crnk-setup/crnk-setup-spring-boot2/src/test/java/io/crnk/spring/metrics/CrnkWebMvcTagsProviderTest.java
```diff
@@ -0,0 +1,78 @@
...
+                "/users/" + id,
+                "/users/{id}"
+            },
+            new Object[] {
+                "/users/" + id + "/attribute",
+                "/users/{id}/attribute"
+            },
+            new Object[] {
+                "/users/" + id + "/relationships/relation",
+                "/users/{id}/relationships/relation"
```
vicmosin: @remmeier as you can see here, these are the URIs go to metrics. I replaced actual value with placeholder to have it more consistent with spring impl. In addition, I am not sure whether crnk can generate other URLs since I have only those in my project. So maybe it should be adjusted

remmeier:
I'm also working in the area the last few weeks, but not yet to Spring Boot specific. Let me have a look at it next week in more detail.

remmeier:
merged as is to do a release, maybe a follow in the next one, I think url handling may can benefit from JsonPath and related classes.

PR #454, opened 2019-01-07
Improved multiple paging resources inclusion

### LINKED ISSUE #453: Weird failure by using multiple pagination
I've found one interesting case which isn't covered by tests and caused failure with:

`java.lang.ClassCastException: io.crnk.core.queryspec.pagingspec.OffsetLimitPagingSpec cannot be cast to io.crnk.core.queryspec.pagingspec.NumberSizePagingSpec`

ResourceA with NumberSizePagingSpec and resourceB with default one (i.e. OffsetLimitPagingSpec).
* As soon as one makes `resourceB?include[resourceA]=...` works fine
* As soon as it makes opposite request i.e. `resourceA?include[resourceB]=...` request fails with error above.

It actually happens because `DefaultQuerySpecUrlMapper` [trying to serialize](https://github.com/crnk-project/crnk-framework/blob/master/crnk-core/src/main/java/io/crnk/core/queryspec/mapper/DefaultQuerySpecUrlMapper.java#L294) the `PagingBehavior` of resourceA with `QuerySpec` holding `PagingBehavior` of resourceB

vicmosin:
Improved logics to retrieve the appropriate PagingSpec from request, based on currently handled ResourceEntry

Fixes #453


vicmosin:
Apparently it fails in other case... Have to review the solution

Issue #462, opened 2019-01-15
Unable to create a resource using a DTO


JochenReinhardt:
When trying to create a resource (POST without ID), I get an error: 
```
java.lang.IllegalStateException: id not available for entity ModernDTO(id=null, value=Test New)
	at io.crnk.core.engine.internal.utils.PreconditionUtil.fail(PreconditionUtil.java:50) ~[crnk-core-2.11.20190113153635.jar:na]
	at io.crnk.core.engine.internal.utils.PreconditionUtil.verify(PreconditionUtil.java:131) ~[crnk-core-2.11.20190113153635.jar:na]
	at io.crnk.jpa.JpaEntityRepositoryBase.saveInternal(JpaEntityRepositoryBase.java:188) ~[crnk-jpa-2.11.20190113153635.jar:na]
	at io.crnk.jpa.JpaEntityRepositoryBase.create(JpaEntityRepositoryBase.java:162) ~[crnk-jpa-2.11.20190113153635.jar:na]
```

I have created a Spring Boot demo application for the issue: https://github.com/JochenReinhardt/crnk-jpa-dto-demo

The issue seems to be with the following lines of code in `JpaEntityRepositoryBase.saveInternal`:
```java
        // id may differ from primary key
        ResourceField idField = getIdField();
        I id = (I) idField.getAccessor().getValue(resource);
        PreconditionUtil.verify(id != null, "id not available for entity %s", resource);
 ```

`resource` is the DTO passed into the `saveInternal` method. The entity is created and an ID is generated automatically by Hibernate. But the incoming `resource` is not updated. This is why the precondition fails in this case.

Is it possible to update the DTO with the ID from the entity in such a case?

After an entity is created or updated, don't we need a call to `em.flush()` to make sure the finder finds the updated entity?

remmeier:
kind of interesting, yes a flush seems necessary to resolve the issue. There is a testSequencePrimaryKey within JpaEntityRepositoryTestBase that covers this particular use case and I have seen it working in other palces. But also noticed it once not working. So something seems of and probably better to add that flush. Altough for bulk updates it might not be desired, so it should be possible to disable it again or maybe even perform the flush only for generated primary keys.

JochenReinhardt:
Most important to me is, that the create works using a DTO.
[...]

PR #463, opened 2019-01-18
Crnk 462 create resource using dto

JochenReinhardt:
As reported in issue #462 the creation of an entity with an autogenerated ID fails when a DTO without ID is passed in.
Created a test case to reproduce the issue and suggested the most simple fix.


crnk-jpa/src/main/java/io/crnk/jpa/JpaEntityRepositoryBase.java
```diff
@@ -185,6 +184,10 @@ protected boolean fetchRelations() {
         // id may differ from primary key
         ResourceField idField = getIdField();
         I id = (I) idField.getAccessor().getValue(resource);
+        // id could have been created during persist
+        if (id == null && pk != null && idField.getElementType().isAssignableFrom(pk.getClass())) {
```
remmeier: oh, I overlooked this one, not related to flush after all. it is a bit tricky in this area. 

I think a safer method would be to check whether primaryKeyAttribute.getName() (on super type) matched the idField.getUnderlyingName. This casae we are safe and can just make use of "pk". If it does not match, the should maybe call a method "getIdFromEntity". This may be overriden by sub types, but by default I would call a mapper.map(entity) and the fetch the ID with getIdField(). not super efficient but correct and ok-ish for a safe method, the effort is spend in other places.

remmeier: it is also not 100% safe if a mapper would have the id of mapping fields with the same name to different other fields, but that seem not to sane. But for the sake of correctness it could make sense to move the entire getIdFromEntity logic into a dedicated method to make it customizable/overridable.

JochenReinhardt: I like the idea of `getIdFromEntity` method.
A few thoughts:
- The precondition check for `pk != null` should not be needed. `pk` cannot be null after a successful call to `em.persist(entity)`. If pk was set, the entity is updated or created. If pk was null it can either be autogenerated or persist will fail. It is impossible to persist an entity without id. So the precondition check will never fail.
- I do not particularly like the idea of calling the mapper. I would have to provide a tuple to the `unmap`. The first entry would have to be the entity, but what about the others? The mapper could rely on the presence of calculated values, etc.

After all, I think my current checks are not too bad.
- The entity's pk is only used if the DTO's id is null
- The only use case is the create use case. In all other cases, the ID must be set.
- The primary key type is checked to be assignable to the resource ID type
- Additional check for matching property names added

remmeier:
thx! lets go with that. I'm not overly fond of doing the mapping either. I guess it is reasonable to assume that people will override this method should they do something overly complex. For 99.9% of the use cases I guess the current solution will always work.

PR #468, opened 2019-02-03
Update TypeTools to 0.6.1 for upstream bugfix

lorenzleutgeb:
Ripple effect of jhalterman/typetools#48


remmeier:
thx, that version should at some point be properly managed to only have to change it once.

PR #471, opened 2019-02-14
WIP: [25] Crnk should respect JsonInclude

a-dudko:
Initial diff in order to implement JsonInclude.NON_NULL (this one only for now) support according to @remmeier advises in the issue's discussion. The tests are currently missing as I'd like to get a feedback first if I'm moving in the right direction and what should be changed/improved/etc.


remmeier:
I sorry for letting you waiting. Looks great already, thx! You already added various tests, could you add one more in the area of DocumentMapperTest? It will test beneath the ResourceMapper where you also had modifications. Add the end of the day this will be the one test that will really check the entire feature.  

In the future I guess more things would be possible, e.g. next to NOT_NULL and DEFAULT also the EMPTY or custom ones.

Issue #483, opened 2019-03-14
Intermittent NullPointerException when creating or deleting resources


sunmit9:
Encountering NPE intermittently when trying to create a new resource with a POST or DELETE api. Following is the log trace from crnk.

`{"exception":{"stacktrace":"java.lang.NullPointerException\n\tat io.crnk.core.repository.foward.strategy.SetOwnerStrategy.addRelations(SetOwnerStrategy.java:64)\n\tat io.crnk.core.repository.foward.ForwardingRelationshipRepository.addRelations(ForwardingRelationshipRepository.java:141)\n\tat io.crnk.core.engine.internal.repository.RelationshipRepositoryAdapterImpl$3.invoke(RelationshipRepositoryAdapterImpl.java:120)\n\tat io.crnk.core.engine.internal.repository.ResponseRepositoryAdapter$RepositoryRequestFilterChainImpl.doFilter(ResponseRepositoryAdapter.java:241)\n\tat io.crnk.core.engine.internal.repository.RelationshipRepositoryAdapterImpl.addRelations(RelationshipRepositoryAdapterImpl.java:129)\n\tat io.crnk.core.engine.internal.dispatcher.controller.FieldResourcePost.attachToParent(FieldResourcePost.java:122)\n\tat io.crnk.core.engine.internal.dispatcher.controller.FieldResourcePost.lambda$handleAsync$2(FieldResourcePost.java:88)\n\tat io.crnk.core.engine.result.ImmediateResult.zipWith(ImmediateResult.java:45)\n\tat io.crnk.core.engine.internal.dispatcher.controller.FieldResourcePost.handleAsync(FieldResourcePost.java:87)\n\tat io.crnk.core.engine.internal.dispatcher.controller.BaseController.handle(BaseController.java:51)","exception_class":"java.lang.NullPointerException"}`

Not sure why the `context.findAll(targetEntry, targetIds, queryContext)` in SetOwnerStrategy.java is returning null.

remmeier:
a sanity check could be added at that place. But I guess that null is directly coming from a repository? a custom implemented one? based on line numbers it is also not a current release?

sunmit9:
[...]

PR #484, opened 2019-03-15
Sanity null check for target relations

sunmit9:
Adding a sanity null check when adding or removing relations.
Issue reported: https://github.com/crnk-project/crnk-framework/issues/483


remmeier:
thx! Not yet sure whether to ignore the null or check and throw an exception (with PreconditionUtil.verify(...)). I'm leaning towards the later. The entire crnk-core part assumes it to not be null. Either way the check could be added to ResourceRepositoryAdapterImpl.findAll(). This way it will be available through the entire code base as everything goes through that adapter.

remmeier:
close due to inactivity, feel free to reopen

PR #507, opened 2019-04-24
Nesting for multiple levels

Mask:
I think it is actually enough (at least for singular nesting) to recursivley call getResourceUrl on the parent to compute the correct url.
I am missing something?

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/engine/internal/registry/ResourceRegistryImpl.java
+++ b/crnk-core/src/main/java/io/crnk/core/engine/internal/registry/ResourceRegistryImpl.java
@@ -162,7 +162,7 @@ public class ResourceRegistryImpl extends ResourceRegistryPartBase implements Re
 			ResourceField childrenField = parentInformation.findRelationshipFieldByName(parentField.getOppositeName());
 
 			if(resourceInformation.isSingularNesting()){
-				String parentUrl = getResourceUrl(queryContext, parentInformation) + "/" + parentInformation.toIdString(id);
+				String parentUrl = getResourceUrl(queryContext, parentInformation, id);
 				return parentUrl + "/" + childrenField.getJsonName();
 			}else{
 				Object parentId = resourceInformation.getParentIdAccessor().getValue(id);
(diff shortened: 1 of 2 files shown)


remmeier:
yes, would be a start. Then there is of course the multi-valued nesting. And to be able to process incoming c alls, the PathBuilder to be able to process incoming calls. 

=> NestedResourceTest could be a good start to extend to verify whether things are working smoothly

Mask:
I'm not quite sure what you mean by extending PathBuilder. 

I have now added CRUD tests based on checkManyCrudWithController() for two-level nested resources, both singular and multiple:

The OneNested Resource now has a nested oneGrandchild Resource, and the ManyNested now have nested manyGrandchildren resources. These grandchildren resources each have "parent" relationships to their owners.

I have one small problem I am having trouble understanding: I had to comment out [this line](https://github.com/Mask/crnk-framework/blob/63c27ea62fbabf4bece216d7fad714efe6ecead7/crnk-core/src/test/java/io/crnk/core/engine/internal/resource/NestedResourceTest.java#L495) in the tests:

When requesting one of the manyGrandchildren resources and including "parent" in the query, I would expect only the direct parent (manyNested) to be included. But for some reason, both the parent and the grandparent are included. This is probably due to the fact that both have a relation named "parent". I think the problem is somewhere around isInclusionRequestedForQueryspec

remmeier:
PathBuilder is responsible for parsing request urls, so if we introduce nesting to multiple layers, it will need to be able to parse this URLs. Otherwise we will only be able to GET nested resource, but not the POST, PATCH, DELETE them. As a disclaimer, PathBuilder is a bit historic and needs a cleanup. So it does not look to pretty :-(

for the second question: please have a look at http://www.crnk.io/releases/stable/documentation/#_properties
=> `crnk.config.include.behavior`

since currently we talk all about crnk 3.0. It might be worthwile to switch the default to `BY_ROOT_PATH`. 

A bit more insight from the JSON:API spec in this area would also be helpful...

Mask:
PathBuilder seems to work as-is, even with mutiple nesting levels. Given a path such as ```/test/b/manyNested/b/manyGrandchildren/c```, the PathBuilder returns ```manyGrandchildren/[b-a-c]```. 

I *did* test POST, PATCH and DELETE requests. Am I missing something? Is there a test I should add to cover some behaviour I have overlooked?

remmeier:
actually looks great, seems to test it quite extensively and the implementation a bit farther along as I had in mind. So lets merge that and refine if we see something comes up.

remmeier:
thx!

PR #512, opened 2019-05-14
Recursive allocateMetaElement only for known Collections (backport for v2.x)

lorenzleutgeb:
We're using Crnk 2.x in a project and were running into issues with `crnk-meta` when resolving self-referential type-declarations such as `S extends Enum<S>`.

This requires:
 1. A fix in typetools, which is currently pending as jhalterman/typetools#53 and should reach Crnk as `net.jodah:typetools:0.6.2`.
 2. A fix in Crnk, which is included in this PR. The problem is that if Crnk *suspects* that a type `T<E>` might be a known collection type, it will recurse to analyze `E`, without checking that `T` actually is some `List` or `Set`. The culprit is, that for structures where `T<E>` is self-referential, and therefore `T` and `E` are indistinguishable, this process will loop and result in a stack overflow. An easy remedy is to check for `T` **before** the recursive call.

Note that these changes can also be integrated into Crnk 3 and I could provide a PR for that. However the system we are developing must go live soon, so from my point of view it is much more important to patch 2.x first. I hope you can understand that, and for the sake of simplicity let's first do 2.x and I will provide a patch for 3.x later on.

### DIFF
--- a/crnk-core/build.gradle
+++ b/crnk-core/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'java'
 
 dependencies {
-	compile 'net.jodah:typetools:0.6.1'
+	compile 'net.jodah:typetools:0.6.2'
 	compile 'org.slf4j:slf4j-api:1.7.25'
 	compile 'com.fasterxml.jackson.core:jackson-databind'
 	compile 'com.fasterxml.jackson.core:jackson-annotations'
--- a/crnk-meta/src/main/java/io/crnk/meta/provider/MetaPartitionBase.java
+++ b/crnk-meta/src/main/java/io/crnk/meta/provider/MetaPartitionBase.java
@@ -151,13 +151,19 @@ public abstract class MetaPartitionBase implements MetaPartition {
 
 		if (type instanceof ParameterizedType && ((ParameterizedType) type).getActualTypeArguments().length == 1) {
 			ParameterizedType paramType = (ParameterizedType) type;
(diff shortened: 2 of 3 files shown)


remmeier:
makes sense, thx! we can merge as soon as type tools is released. 

potentially in crnk 3 we want to also a support for the base MetaCollectionType, not just list/set.

lorenzleutgeb:
Great. Thanks a lot for the fast feedback!

remmeier:
I tried again to restart the build, but so far the PR is still failing  with:

```
* What went wrong:
Execution failed for task ':crnk-core:compileJava'.
> Could not resolve all files for configuration ':crnk-core:compileClasspath'.
   > Could not find net.jodah:typetools:0.6.2.
     Required by:
         project :crnk-core
```

lorenzleutgeb:
Yes, that's because jhalterman/typetools#53 is still pending, I am waiting for @jhalterman's response.

lorenzleutgeb:
TypeTools 0.6.2 has been released, so this PR should be reconsidered. I wanted to retry the build on Travis CI, but apparently that's not possible (maybe because it is too old?!). @remmeier please advise how to proceed. Thank you!

lorenzleutgeb:
gg

remmeier:
maybe can you try to createa new one? not really sure what the issue with travis is.

lorenzleutgeb:
All checks have passed now (see GitHub status indication below the last comment).

Issue #536, opened 2019-06-28
Jackson JsonAnySetter support to for dynamic fields using crnk-client


magg:
Hello 

According to the documentation JsonAnySetter and JsonAnyGetter are supported but my tests show otherwise

I'm using ` <crnk.version>3.0.20190606072637-BETA</crnk.version>`

I'm trying to create an object and the attributes fields are not included on creation. I'm using to crnk-client to create a object on a remote server. Funny thing is, the json it's populated correctly when it comes back from the server. But it's not populated correctly on the client before sending it to the server

```
@JsonApiResource
@JsonInclude(JsonInclude.Include.NON_DEFAULT)
public class Customer {

    @JsonApiId
    protected Long id;
    private String name;


    @JsonIgnore
    protected Map<String, Object> customFields = new LinkedHashMap<>();

    @JsonAnyGetter
    public Map<String,Object> any() {
        return customFields;
    }

    @JsonAnySetter
    public void set(String name, Object value) {
        customFields.put(name, value);
    }
```

it generates the following: (all fields in the customFields map are missing from the attributes json block)

```
"data” : {
   “type” : “customer”,
   “attributes” : {
     “name” : “Amy”
   }
 }
```

I'm instantiating the class like so Customer c = new Customer(); and setting c.set("email", "lol@outlook.com"); c.setName("Amy")

Any help is appreciated

thanks!
- Miguel

remmeier:
looks like it is not supported as of yet in the client. Historically the @JsonApiSetter has been much older than crnk-client and not so much used. So it seems that it slipped.

There is a `ResourceInformation.getAnyFieldAccessor` that needs to be adopted in `ResourceMapper.setAttributes`. Then it shoulds be fine.

magg:
hi @remmeier is it an easy change then? do you think it can be merged soon?

remmeier:
yes rather small I think. If you have time you can have a look. Otherwise I try to make time.

magg:
Hello @remmeier can you please review PR https://github.com/crnk-project/crnk-framework/pull/537

I tested and it it generates the json i need, I need this to be merged before Monday hopefully

magg:
closing issue as my PR was already merged

thanks!

PR #537, opened 2019-06-29
fix for issue 536: Jackson JsonAnySetter support to for dynamic fields using crnk-client

magg:
Please see issue https://github.com/crnk-project/crnk-framework/issues/536


crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/ResourceMapper.java
```diff
@@ -123,7 +126,18 @@ protected void setAttributes(Resource resource, Object entity, ResourceInformati
 					setAttribute(resource, field, value);
 				}
 			}
+
+		}
+
+		if (resourceInformation.getAnyFieldAccessor() != null) {
+			AnyResourceFieldAccessor anyFieldAccessor = resourceInformation.getAnyFieldAccessor();
+			Object o = anyFieldAccessor.getValues(entity);
```
remmeier: The anyFieldAccessor could return a Map directly I guess?

remmeier:
looks great, thx!  added one small comment. Potentially a one small test case on the client side would be good to make sure it remains working. For example in BasicRepositoryAccessTestBase that is inherited from RepositoryAccessClientTest and adding the any getter/setter to one of the types like Schedule..

crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/ResourceMapper.java
magg: OK I will update it, sound good

magg:
Hi @remmeier I changed the code according to your comment and added the test, but it's failing, could you please take a look?

magg:
so there's a weird bug... when run this test it fails
```
    @Test
    public void testCreateAnyFields() {
        Schedule schedule = new Schedule();
        schedule.setName("mySchedule");
        schedule.setAnyFields("randomField1", 1234);
        schedule.setAnyFields("randomField2", "test");

        scheduleRepo.create(schedule);

        QuerySpec querySpec = new QuerySpec(Schedule.class);
        ResourceList<Schedule> list = scheduleRepo.findAll(querySpec);
        Assert.assertEquals(1, list.size());
        schedule = list.get(0);
        Assert.assertNotNull(schedule.getId());
        Assert.assertEquals(1234, schedule.getAnyFields().get("randomField1"));
        Assert.assertEquals("test", schedule.getAnyFields().get("randomField2"));

    }

```


scheduleRepo.findAll(querySpec) does not bring the extra fields, but scheduleRepo.create(schedule) does return them. I committed the code where i check scheduleRepo.create(schedule) output

Do you have any ideas if there's a difference in the handling of create and findAll? ResourceUpsert.setAttribute does the handling in the back-end right?

magg:
So @remmeier I think the problem with the findOne/findAll methods rely on DefaultQueryPathResolver class, it has a variable allowUnknownAttributes = false;

I found a workaround to fix the findOne/findAll methods, 

in my code this fixed the issue by setting a new DefaultQuerySpecUrlMapper in the CrnkClient

```       
DefaultQuerySpecUrlMapper urlMapper = new DefaultQuerySpecUrlMapper();
urlMapper.setAllowUnknownAttributes(true);
client.setUrlMapper(urlMapper);
```

But I think this cannot be added in BasicRepositoryAccessTestBase without it breaking many tests.

remmeier:
ResourceUpset does the update processing logic, whereas ResourceMapper translates objects to json structures. So they work in opposite directions but maybe could be merged at some point for clarity.

allowUnknownAttributes is only involved for urls (sorting, filtering). Ideally it would back-off automatically is the resourceInformation has any setters. 

I can take a look this evening if it does not clear up.

magg:
added a simple check in DefaultQueryPathResolver.java to check jsonanysetter too

`if (allowUnknownAttributes || anyFieldAccessor != null) `

remmeier:
thx a lot!

Issue #532, opened 2019-06-24
AsciidocCaptureModule wrong decorate ManyRelationshipRepository


0ffer:
When I am try to get ManyRelationshipRepository from CrnkClient that wrapped with AsciidocCaptureModule I get an error 
```
java.lang.ClassCastException: class io.crnk.gen.asciidoc.capture.AsciidocCaptureModule$AsciidocOneRelationshipDecorator cannot be cast to class io.crnk.core.repository.ManyRelationshipRepository (io.crnk.gen.asciidoc.capture.AsciidocCaptureModule$AsciidocOneRelationshipDecorator and io.crnk.core.repository.ManyRelationshipRepository are in unnamed module of loader 'app')
```

I found, that AsciidocCaptureModule wraps repositories not correctly.
The module contains the next code:
```java
@Override
        public Object decorateRepository(Object repository) {
            if (repository instanceof ResourceRepository) {
                return new AsciidocResourceDecorator((ResourceRepository) repository);
            }
            if (repository instanceof OneRelationshipRepository) {
                return new AsciidocOneRelationshipDecorator((OneRelationshipRepository) repository);
            }
            if (repository instanceof ManyRelationshipRepository) {
                return new AsciidocManyRelationshipDecorator((ManyRelationshipRepository) repository);
            }
            return repository;
        }
``` 
And when it runs the next repository type come in:

```java
@Deprecated
public interface RelationshipRepository<T, I, D, J>
        extends Repository, MatchedRelationshipRepository,
        ManyRelationshipRepository<T, I, D, J>, OneRelationshipRepository<T, I, D, J>
```

Now it is impossible to get correctly decorated ManyRelationshipRepository because if-statement always stopped on OneRelationshipRepository.

remmeier:
I will do the 3.0 release these days. Should you have time to look at  this, let me know. Or otherwise I can take a quick look before the release.

0ffer:
> I will do the 3.0 release these days. Should you have time to look at this, let me know. Or otherwise I can take a quick look before the release.

Yes, I will try to fix it. I already forked repository and saw the problem code.

0ffer:
Fixed with PR#538

PR #538, opened 2019-07-01
#532: AsciidocCaptureModule wrong decorate ManyRelationshipRepository

0ffer:
Add new decorator class for RelationshipRepository entity and redirect calss to one and many decorators.
Add some tests for check decoration correctness.

@remmeier, please see changes. I am not sure that I do right work with the problem.

### DIFF
--- a/crnk-gen/crnk-gen-asciidoc/src/main/java/io/crnk/gen/asciidoc/capture/AsciidocCaptureModule.java
+++ b/crnk-gen/crnk-gen-asciidoc/src/main/java/io/crnk/gen/asciidoc/capture/AsciidocCaptureModule.java
@@ -1,10 +1,5 @@
 package io.crnk.gen.asciidoc.capture;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.Collection;
-import java.util.Map;
-
 import io.crnk.client.http.HttpAdapter;
 import io.crnk.client.http.HttpAdapterListener;
 import io.crnk.client.http.HttpAdapterRequest;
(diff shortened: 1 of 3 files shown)


crnk-gen/crnk-gen-asciidoc/src/main/java/io/crnk/gen/asciidoc/capture/AsciidocCaptureModule.java
```diff
@@ -290,6 +291,58 @@ public void delete(I id) {
 		}
 	}
 
+	@Deprecated
```
remmeier: you can remove the deprecator, it is perfectly valid to implement both oneRelationship repository and manyRelationship repository. Looks good otherwise. Maybe it would make sense to have this combining one, many wrapper in crnk-core next to WrappedRelationshipRepository as the use case comes up in other places.

0ffer: Thanks. I made some changes and move combine wrapper to package with wrappers.

0ffer:
@remmeier please look at changes. I made changes

crnk-gen/crnk-gen-asciidoc/src/main/java/io/crnk/gen/asciidoc/capture/AsciidocCaptureModule.java
```diff
@@ -95,15 +90,32 @@ public Object decorateRepository(Object repository) {
 			if (repository instanceof ResourceRepository) {
 				return new AsciidocResourceDecorator((ResourceRepository) repository, this);
 			}
+			if (repository instanceof RelationshipRepository) {
```
remmeier: looks great like this! 

One small last thing  that we can merge: can you change `repository instanceOf RelationshipRepository` to `repository instanceof OneRelationshipRepository && repository instanceof ManYRelationshipRepository`?

remmeier: because the old RelationshipRepository is deprecated, but users may still continue to implement both one and many in the same class.

0ffer: Ok! In several hours I will make the changes.

0ffer:
I commit your suggestion. You can look at them.

remmeier:
perfect, thx!

Issue #539, opened 2019-07-01
using dataroom filter with reactive repositories


st-h:
I started to have a look at Crnk because I need json-api support for vertx. So far I have been able to figure out everything until I started to try out dataroom. Crnk logs warnings for all my registered repositories:

```
SecurityModule - no dataroom support for unknown repository xyz
```

I had a look at the [implementation](https://github.com/crnk-project/crnk-framework/blob/ff78f9d1a10b0d5e4f260f095bb1f5d6b8e8f4ca/crnk-security/src/main/java/io/crnk/security/SecurityModule.java#L201) and it seems that there currently is no support for reactive repositories yet since `ReactiveResourceRepository` does not inherit from `ResourceRepository`?

I'll be happy to take a stab at contributing a PR if we continue to use crnk, but most likely I will need a few pointers about what should be done as I am currently just starting  to look into Crnk.

remmeier:
yes, true. For the basics most things are already in place. One just has to setup a reactive counter-part to the DataRoomResourceFilter. Potentially one may want to have a ReactiveDataRoomFilter as well to make the computation of the filter reactive.

st-h:
@remmeier thanks for the reply. Does not sound like its very complicated to implement. However, so far I have been running into issues when integrating crnk with the vertx auth handlers, specifically together with vertx-web. The problems seems that the crnk `SecurityProvider` needs to be set up at a module level, however the `user` that vertx provides is only available at the `request` or `routingContext` level. Do you know if anyone has already made some plans regarding auth integration with vertx? It kind of looks to me that the concepts from Vertx-web and crnk will not work together nicely, but I might still be missing something fundamental. One could use vertx without the web extension, however that would also mean that a lot of what vertx-web provides will not be available (things like handling session data or cookies for instance)

remmeier:
There are similar aspects with the spring, servlet, jaxrs, etc. integration. Certain things are only available from an outside context since we don't want to drag along all things in our repository (and other) contracts. For non-reactive systems this usually then ends up with thread- and request-scoped variables with the later being a automatically garbage-collected thread-scope as well. It is one of the benefit in the crnk architecture I think. Because client and servers are symmetric, can make use of the same repository contracts. context-related things are implement as independent, orthogonal concern.

[...]

PR #541, opened 2019-07-02
WIP: vertx integration

st-h:


crnk-core/src/main/java/io/crnk/core/engine/security/SecurityProvider.java
```diff
@@ -1,8 +1,10 @@
 package io.crnk.core.engine.security;
 
+import io.crnk.core.engine.result.Result;
+
 public interface SecurityProvider {
 
-	boolean isUserInRole(String role);
+	Result<Boolean> isUserInRole(String role);
```
remmeier: based on https://github.com/crnk-project/crnk-framework/issues/539#issuecomment-507581394 maybe we should rather have:

```
public interface SecurityProvider {
   boolean isUserInRole(SecurityProviderContext context, String role);
}

interface SecurityProviderContext {
   QueryContext getQueryContext();
}
```

issue is that all security related code is called quite extensively. So it must return immediately.

st-h: alright. Reverted previous change and added the SecurityProviderContext.

remmeier:
looks good like this I think? next question would be whether we can now populate that context with all callers?

st-h:
@remmeier dang, I don't think I really understood your actual plan 😞 
Maybe having a look at it from the other side might help. For vertx we currently have a handler which attaches to vertx like this:
```
server.requestStream().toFlowable()
				.flatMap(request -> handler.process(request))
				.subscribe((response) -> LOGGER.debug("delivered response {}", response), error -> LOGGER.debug("error occured", error));
```
That is essentially just taking the vertx `HttpServerRequest` and passing it to the `CrnkVertxHandler`

As for vertx web I have been changing this to:
```
router.route().handler(routingContext -> {
        HttpServerRequest request = routingContext.request();
        User user  = routingContext.user();
        
        Flowable.fromArray(request).flatMap(r -> handler.process(r))
          .subscribe((response) -> LOGGER.debug("delivered response {}", response), error -> LOGGER.debug("error occurred", error));

      });
```
Hopefully, there is a more elegant way to patch things together, but I am pretty much doing the same thing: Passing the `vertx` request to the `crnk` handler. However, in this case we also have a `user` object which comes along with the request within the `routingContext`. I think we now would need a way to pass that user along with the request, so it eventually will end up within the `DataRoomFilter` and we can check wether that user is allowed to make the request. Does that sound right? The `CrnkVertxHandler` passes the `VertxRequestContext`/`DefaultHttpRequestContextBase`/`HttpRequestContextBase` to the dispatcher. Either we can add the user somewhere here or we would need another concept (I didn't grasp the concept of using Mono.attachContext fully, as it seems we will need to also pass along the Result so we can access what we tried to store? or does this automatically happen by attaching to some thread in the background?)

I am still having trouble figuring out where we can attach that user to, as we cannot just use the thread based approach which can be found within the JaxrsModule for instance.

st-h:
Well, the longer I am thinking about this, the more I feel intrigued by the idea that allowing to add custom data per request to the `querySpec` would resolve most of the issues I am currently facing. This would allow to add data concerning authentication which could be used:
- by repositories and would allow to bypass crnks authentication support and leave it up to be handled by business logic (unfortunately this currently would only work for find methods)
- by dataroomFilters to determine if the resource may be accessed

probably we would still need a way to make the same data available to the SecurityProvider so this will work for ResourcePermissions etc as well. What do you think?

However, by now I am getting close to dropping crnk altogether. I have been looking through the source and the docs for quite some time now and I still seem to not be able to grasp the whole authentication/authorization concept, so all this has become a little frustrating by now :( Yet, I quite like the support of different repositories and how Mono allows to drop into most reactive business logic.

remmeier:
QueryContext.attributes would be the place where one can add arbitrary things like the user object. I can also finish up this PR. It goes a bit deep into a lot of different details which makes it rather hard to do. I would have time for this. 

> However, by now I am getting close to dropping crnk altogethe

support for vert.x is indeed still "rudimentary". But you can also make a list of what you would like to see and we can make it happen. It is not overly much effort when knowing things. Since there are so many frameworks out there, things are a bit "demand-driven".

I would not recommend py-passing the built-in security mechanisms. You can of course do that, but you also loose a number of things. For example, it a user is not authorized to see one resource, all the relationships will be removed as well.

remmeier:
one consideration is to add QueryContext as attribute to QuerySpec. Not overly fond of it. But could provide a shortcut from QuerySpec deeper into the system without having to worry about subscriberContext and stuff.

st-h:
@remmeier thanks a lot for your help and support. One thing I have been wondering about are situations where it might be difficult to provide all relevant data to the LoginModule upfront before querying the repositories. For instance permissions might differ if the current user is the owner (or has created) a parent object (the id of that object is provided within the request). So, essentially one needs to retrieve records form the database before being able to decide if the request should be fulfilled. The problem I see is that we can get the current user from within the vertx requestContext (before we pass the request to crnk). However, at this point we do not know about the actual request params etc, that crnk will extract later. At a later stage (like in DataRoomFilter or the repositories) the request params are available, but currently I could not find a way to get the current user. So in both places only a subset of the required data is available.

> QueryContext.attributes would be the place where one can add arbitrary things like the user object. I can also finish up this PR. It goes a bit deep into a lot of different details which makes it rather hard to do. I would have time for this.

I actually think that probably will be much more efficient. I still haven't fully grasped how the changes I made in this PR will fit in with what I am trying to achieve. I probably will still need to ask a lot of questions to get this finished.

> one consideration is to add QueryContext as attribute to QuerySpec. Not overly fond of it. But could provide a shortcut from QuerySpec deeper into the system without having to worry about subscriberContext and stuff.

that's just an idea I was having to solve the issues I am currently facing - but I certainly can't tell if there would be better ways to do so, as I am still struggling to grasp all the details about crnk internals.  But I think that at least sometimes there will be situations where one is not able to fully separate things into (I) stuff that needs to happen before crnk (like getting the current user and its permissions) and (II) the actual data retrieval in repositories which only relies on request params. I think a way of making data available within repositories etc that we had before we called the crnk handler would be very helpful.

I would like to keep adopting crnk and JsonApi on our roadmap, but for now I am afraid I will have to stick with replacing the web socket with a rest interface and keep our custom data format, as I sadly can not really make up for a lot more time right now.

remmeier:
does vertx have a role-based access control concept? or just a user name?

I will take care of this PR this week (v3 cleanup is should be able to finish today/tomorrow). And you can look at things in due time. vert.x insights and tickets would be great to bring it forward in that area should you have time.

st-h:
@remmeier vertx itself doesn't. The web module has pluggable [authentication/authorization support](https://vertx.io/docs/vertx-web/groovy/#_authentication_authorisation) where some [modules](https://vertx.io/docs/#authentication_and_authorisation) with varying features are provided.
The [user](https://vertx.io/docs/apidocs/io/vertx/ext/auth/AbstractUser.html) class all this is based on comes with a method to check permissions, so I think the short answer is yes.

I'll try to follow along your changes, but I only have been starting to replace older custom implementations with features from vertx-web recently and the docs are a little sparse on best practices, so I am not always sure if I am choosing the best approach.

By now I am also wondering if some of the issues I was having, are more related to our general use cases than the vertx support, which is more of a drop-box like web application. So, we can not just make simple decisions regarding authorisation based on resource types alone, but by the actual data of the user and other resources. Maybe the only limitation actually is that crnk expects to be able to look up the user/principal from a context that is tied to the current thread, which vertx due to its non blocking nature does not provide. However I am still not sure how one would implement a data based policy when there does not seem to be a way to get the current user from within a DataRoomFilter (as neither the QuerySpec nor the SecurityProvider expose any go that information) - and there is no way to get the current user from within a repository to sort it out later when the data of the resource is known. To sum it up very shortly, I am still not sure if this a general crnk or a vertx support issue or if I just have been missing something very important.

remmeier:
I rebased and extended this commit in origin/vertx. I finished up the refactoring of SecurityProvider. The interesting piece is VertxSecurityProvider. It makes use of the `QueryContext` (which I my rename to `CallerContext` or something similar).  It is hooked into the system by `CrnkVerticle`, both as security module and as interceptor early in the request processing to do authentication and put it into `QueryContext`.

remmeier:
https://github.com/crnk-project/crnk-framework/tree/vertx

remmeier:
missing so far is the real role-based access control integration and adding support for reactive repositories by dataroom filtering. But those are quite close as well now.

remmeier:
closing this one in favor of https://github.com/crnk-project/crnk-framework/pull/546 where I continued working on it. Exampe application already integrates the security module now. no reactive dataroom repositories yet.

PR #546, opened 2019-07-09
support SecurityModule for reactive repositories #539

remmeier:


remmeier:
add first support to master branch: follow up to add further dataroom support (reactive repositories missing as of yet)

PR #553, opened 2019-07-24
ClassUtils getRawType function should check for WildcardType

jbmolle:
Hi,
I'm using Crnk with Kotlin. I've created a User model and assign it a set of Role. The signature of Set in Kotlin is Set<out E> and Role is a Java class so not final by default. 
At runtime it gives the following signatrue Set<? extends io.crnk.security.repository.Role>.
? extends io.crnk.security.repository.Role is a WildcardType and in ClassUtils getRawType it will throw an IllegalStateException with unknown type.
This pull request add the required conditional branch for WildcardType and solve the bug.
Best,
JB


remmeier:
thx! makes absolutetly sense. That one failing test is not related to this PR. It seems to happen randomly at some times. I will have a look at it and merge this one.

Issue #506, opened 2019-04-24
Force inclusion of all relationships in OperationsModule when performing multiple operations


duncanportelli:
I am using Crnk's  `OperationsModule` and am performing multiple (i.e. more than one) POST operations. However, after successfully carrying out all POST operations, this module is re-retrieving all resources which were posted and [forcibly including all relationships](https://github.com/crnk-project/crnk-framework/blob/70b1d99f6c55820ac65d929ed4b9bc096c966752/crnk-operations/src/main/java/io/crnk/operations/server/OperationsModule.java#L209) in the request. 

I guess this is done in order to retrieve the latest data after all the operations have been performed. However, my question is, why are ALL the relationships being retrieved?

remmeier:
I guess it should become configurable, disabled by default or further optimized where @JsonApiRelationId is applied. Main motivation in the past was at if something gets patched, the result will show you the patched document in full, including the relationship ids.

brandonandre:
As a workaround to help speed up your operations POST request, you can apply the `crnk-compact: true` header with the request. I get a noticeable speed increase with responses.

PR #556, opened 2019-08-02
Operations Including All Relationships Toggle

brandonandre:
Added a couple of toggles in the operations module to fix issues in #506. This greatly improves performance for entities which contain many foreign keys. Added a displayOperationResponseOnSuccess toggle which allows to only display the status to decrease the size of the response you get back. 

Also updated the documentation to reflect these changes. 

Thanks!

### DIFF
--- a/crnk-documentation/src/docs/asciidoc/operations.adoc
+++ b/crnk-documentation/src/docs/asciidoc/operations.adoc
@@ -40,6 +40,30 @@ There is further an operations client implementation that works along the regula
 include::../../../../crnk-operations/src/test/java/io/crnk/operations/OperationsPostTest.java[tags=client]
 ----
 
+## Operation Module Properties
+
+The operations module can be customized using one of the following properties:
+
+.Operation Module Properties
+|===
+|*Property Name* |*Description* |*Default Value*
+|resumeOnError | If an error occurs during the operation execution, continue the other operations. | False
+|includeAllRelationships | Toggle to include all relationships in the request by default. | True
+|displayResultsOnSuccess | If false, it will only display the data responses if an error has occured. Status code are still shown either way. | True
+|===
+
+Here is an example of how to set these properties:
+
+[source]
+----
+OperationsModule operationsModule = OperationsModule.create();
+operationsModule.setIncludeAllRelationships(false);
+operationsModule.resumeOnError(true);
+...
+----
+
+## Example Request and Response
+
 An example request in JSON looks like:
 
 [source]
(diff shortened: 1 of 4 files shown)


remmeier:
thank you! will have a look on sunday/monday.

crnk-documentation/src/docs/asciidoc/operations.adoc
```diff
@@ -40,6 +40,30 @@ There is further an operations client implementation that works along the regula
...
 
+## Operation Module Properties
+
+The operations module can be customized using one of the following properties:
+
+.Operation Module Properties
+|===
+|*Property Name* |*Description* |*Default Value*
+|resumeOnError | If an error occurs during the operation execution, continue the other operations. | False
+|includeAllRelationships | Toggle to include all relationships in the request by default. | True
```
remmeier: naming of this property should change I think. It is more a "includeChangedRelationships". So everything that gets POSTED/PATCHED will be returned as such.

Looking at it, it would be good to update the implementation of that feature (independent of this PR). Currently it triggers a full inclusion, it rather should trigger SerializeType.ONLY_ID.

remmeier:
sorry for the delay. Looks awesome! I left one small note.

crnk-documentation/src/docs/asciidoc/operations.adoc
brandonandre: Hi @remmeier,

I'm also thinking more about this now, and I'm confused on the purpose of the include parameter on this line:

`parameters.put("include", getLoadedRelationshipNames(resource));`

When actually running the operation with a bunch of POST requests, it will return a response with "self" and "related" links for each of the relationships. And the response does not have a "included" section to it. So it seems in my case, it's running the SQL for the include, but not actually displaying the include in the response back.

It seems that the fetchUpToDateResponses method is performing the include but not displaying the results of the include. 

Perhaps the "include" parameter should just be removed all together? Or I might be misunderstanding the purpose of it.

Thanks in advance!

remmeier: it currently triggers an "inclusion" to get hands on the IDs and throwing everything else away. So the results will have relationship/data-elements holding the IDs of POSTed/PATCHed resources.

I created https://github.com/crnk-project/crnk-framework/issues/559 to address it. It should not do any relationship loading, but if possible still display the relationship IDs even if it is not EAGER/ID_ONLY.

brandonandre: Ah I see, that makes sense. Okay, I will change the name of the toggle as suggested above and it should be ready soon.

Thank you for explaining that

remmeier:
thx!

PR #561, opened 2019-08-10
Innapropriate 'not a JSON-API request' when operations module is used…

remmeier:
due to content type application/json-patch+json => moved this check to a later stage and only printed if a repository is found, accept header matches but content type is wrong. Most likely scenario is then that it was an invalid content type.


birdli66:
Quite a change for the minor issue (may be github should highlight tab-to-spaces changes ;-).
Fix looks good to me, many thanks.

PR #572, opened 2019-09-23
Fixed problems with crnk-integration-examples builds

bsshashank:
Fixed gradle build problems with running the example projects under crnk-integration-example.


remmeier:
thx!

PR #575, opened 2019-09-24
Multi-Level Nested Resources (2nd Try)

Mask:
- getResourceUrl() now supportes multi-level nested resources
- added a getResourcePath() in case the caller doesn't want the full URL
- added support to OperationsModule for nested resources


remmeier:
thx again :-)

Issue #237, opened 2018-02-22
Create crnk-gen-openapi to generate Swagger/OpenAPI definition


remmeier:
out of repositories. As a Swagger setup seems quite common/expected.

remmeier:
could work quite similar to typescript generation, may even reuse the the gradle plugin

julillosamaral:
I'm trying to do this. I don't know how much I will be able to advance and to integrate with crnk but at least I'll have a gist generating the swagger doc that can be integrated into crnk after.

The approach I'm doing is generate the doc using swagger-core (https://github.com/swagger-api/swagger-core). Any comments about this?

remmeier:
that would be fairly awesome if you have time to get this done.

What would be necessary is to generate the swagger doc out of a crnk/JSONAPI endpoint. The problem with the swagger/JSON API combination is that JSON API already specifies tons of things which makes directly writing swagger documentation quite time consuming/uncessary. For JSON API all that would actually be needed is to specify the resources and relationships (much simpler than specifying the entire rest endpoint). Based on generated swagger doc clients may then do generation with swagger-core to access the JSON API endpoint.

similar to gen-typescript, crnk-meta can be the basis to perform the generation. Not sure whether swagger-core provides something to easily write a swagger doc file. To some degree the approach is similar to generating the swagger doc out of JAXRS/swagger annotations.

at some point we can refactor the typescript-generator plugin to a more generic plugin where typescript/swagger can be plugged in.

remmeier:
(can help along the way)

julillosamaral:
swagger-core specifies a set of classes to create the specification. My team and I have some experience on swagger specification but we generate it using PHP, doing it in Java will take me a bit more of time but I'm sure that we can do it.

For the moment I think the most difficult part is to get the information from crnk but I think I'll be able to do it. I'll let you know if I don't find a way to do something.

I'll check the typescript generator a bit more but as far as I understood it used some templates (am I wrong?). The approach I'm doing with swagger-core is a bit different, I'm instantiating all the classes and after that swagger-core has a method to dump those objects created into a JSON.

Regards,
Julio

remmeier:
typescript generation takes MetaResource, MetaAttribute, etc. from crnk-meta and transforms it into a typescript meta model (TSInterface, etc.). Afterwards it is written to files with TSWriter.  So it may not be too unsimilar.

remmeier:
[...]

PR #581, opened 2019-10-02
Adds OpenApi generation module crnk-gen-openapi

macleodbroad-wf:
Closes https://github.com/crnk-project/crnk-framework/issues/237

This introduces an **experimental** crnk-gen-openapi module for generating OpenApi v3 specification from crnk projects.

Generated openapi.yaml output can be inspected in the online swagger editor: http://editor.swagger.io/

The **experimental** label will come off as the generator reaches maturity. Currently the known remaining work includes:

- Support for more overrides during build (All Info / Servers)
- Ensuring that when starting with a template, all template data is merged onto the generated code (currently only has partial support)
- Support for openapi annotations to override generated code and add constraints/patterns on resource attributes
- Support for the Plain JSON flavour of crnk
- Support for Bulk Operations / JSON Patch
- Support for json output and other arguments to the output format / options
- csv parameters should set the `style` attribute to something like:
```
style:
	format: "form"
	explode: false
```
- fields[<associated_resource_type>] for every resource that has associated resources
- Support for Delete request bodies (Currently OpenAPI does not allow DELETE methods to define a RequestBody - https://github.com/OAI/OpenAPI-Specification/issues/1801)
- Make use of re-usable parameter groups when https://github.com/OAI/OpenAPI-Specification/issues/445 lands


remmeier:
Thx a lot for that very substantial PR! :-)

PR #622, opened 2019-11-10
ErrorResponse.getResponse() serializes as "entity" instead of "errors"

mitcoding:
Not sure if this is by design, so ignore this pull request if I'm in error. not sure if this was a design decision or an oversight, as the git history of the ErrorResponse method getResponse() from day one has always been returning "new JsonApiResponse().setEntity(data);" instead of "new JsonApiResponse().setErrors(data);". 

At the same time even though it has always been calling setEntity, I think this might be an oversight since the online [spec](https://jsonapi.org/format/#error-objects) currently states: "Error objects MUST be returned as an array keyed by errors in the top level of a JSON:API document."


remmeier:
thx! you shocked me for a short moment... :-) but in general all this fine:

grafik

it does not look like the method is used at all and is rather the `toResponse` method that is used for real system. Maybe this one should be deprecated.

PR #632, opened 2019-11-20
Make MetaResourcePartition respect the access properties of JsonApiResource

ralphmarchildon-wf:
The creation of new `MetaResource` items was *not* using the access values in the incoming `ResourceInformation` source element, defaulting to "true" for whether a MetaResource item could be created, read, updated or deleted, even if specified otherwise.

This lead to the OpenAPI spec over-generating paths for operations that were not valid.

@macleodbroad-wf 
@remmeier


crnk-gen/crnk-gen-openapi/src/test/java/io/crnk/gen/openapi/OpenAPIGeneratorTestBase.java
```diff
@@ -72,9 +72,10 @@ void compare(String expectedSourceFileName, String actualSourcePath) throws IOEx
   }
 
   private static void assertOperationResponseCodes(Operation operation, List<String> codes) {
-  	if(operation == null){
-  		throw new IllegalStateException();
-	}
+    // Some operations will be null in cases where a JsonApiResource has postable, patchable,
+    // readable, deleteable set to false. In that case, do not perform any response checks
+    if(operation == null) { return; }
```
macleodbroad-wf: Should the null operations even be getting this far?

ralphmarchildon-wf: In order to avoid a bunch of `null` checks sprinkled everywhere else, I thought it best to skip here. The original assumption was that a `null` Operation would be invalid, when in fact there *should* be null operations when that operation isn't permitted.

image

macleodbroad-wf: I see

macleodbroad-wf: Should the paths even exist?

ralphmarchildon-wf: Paths? Yes, one or more of the individual Operations on the path might not though. 

Example: something that is immutable in a system may allow `GET /thing/{ThingId}` and `DELETE /thing/{ThingId}` but not `PATCH /thing/{ThingId}`. In the screenshot above, we'd want to skip PATCH response validation because `getPatch()` returns null.

macleodbroad-wf:
+1

remmeier:
thx!

PR #635, opened 2019-11-21
Bulk Patch Spike

macleodbroad-wf:
This is a spike to start discussion on what else is missing beyond openapi generation support, in order to do Bulk PATCH

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/engine/internal/dispatcher/controller/ResourcePatchController.java
+++ b/crnk-core/src/main/java/io/crnk/core/engine/internal/dispatcher/controller/ResourcePatchController.java
@@ -1,15 +1,5 @@
 package io.crnk.core.engine.internal.dispatcher.controller;
 
-import java.io.IOException;
-import java.io.Serializable;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.Callable;
-
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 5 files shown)


crnk-core/src/main/java/io/crnk/core/engine/internal/dispatcher/controller/ResourcePatchController.java
```diff
@@ -42,40 +44,54 @@ protected HttpMethod getHttpMethod() {
...
+			else {
+			// TODO: Merge this instead of creating new
+				
+				Object entity = newEntity(resourceInformation, resourceBody);
+				setId(resourceBody, entity, resourceInformation);
+				setType(resourceBody, entity);
+				setAttributes(resourceBody, entity, resourceInformation, queryContext);
+				setMeta(resourceBody, entity, resourceInformation);
+				setLinks(resourceBody, entity, resourceInformation);
+				Result zipped = setRelationsAsync(entity, registryEntry, resourceBody, queryAdapter, false);
```
remmeier: for truely bulk support we would like to also bulk-enable setRelationsAsync

crnk-core/src/main/java/io/crnk/core/engine/internal/dispatcher/controller/ResourcePatchController.java
```diff
@@ -42,40 +44,54 @@ protected HttpMethod getHttpMethod() {
 
 	@Override
 	public boolean isAcceptable(JsonPath jsonPath, String method) {
```
remmeier: CrnkClient will also need to modify slightly in the ResourceRepositoryStubImpl. Because typically the semantics is /api/xy/{id} but with bulk that {id} is missing. I assume a list of ids is not really  (or having a list of ids?, but I think rather not).

PR #643, opened 2019-12-04
Adds ProxyResource interface to support consuming a remote resource without (complete) knowledge of its structure 

mikenightingale:


remmeier:
Hi! also have been thinking about more support in this area. But in my case more for relationships later than attributes, like proxying one resource without its relationship and still having it working. 

Currently `@JsonAnySetter` and `@JsonAnyGetter` allow to achieve the same functionality as `ProxyResource`, but through annotations rather than an interface (with its advantages and disadvantages).  Neither covers relationships. One can use the low-level `Resource` class to have everything fully dynamic. I'm not yet quite sure in which direction to move. Would you have more plans for `ProxyResource` over  `@JsonAnySetter` and `@JsonAnyGetter`? Kind of related to this I recently introduced `ResourceTypeHolder` which holds the type rather than the attributes, but otherwise serves the same purpose.

mikenightingale:
Thanks for the feedback. As you suggest, @JsonAnySetter and @JsonAnyGetter solves my issue. I've reworked my example and my test pass. Happy to close the PR.

Issue #636, opened 2019-11-25
getQuerySpec fails for different endpoints with same resource class


jbmolle:
Hi,
I have 2 different resources with the same resource class:
- Test with resource class DynamicObject.java
- TestRelation with resource class DynamicObject.java

Test has one relation testRelation of type TestRelation.
When I list all the Tests and include testRelation, the call fails.

It comes from this part of code:
public QuerySpec getOrCreateQuerySpec(Class<?> targetResourceClass, String targetResourceType) {
        verifyNotNull(targetResourceClass, targetResourceType);

        QuerySpec querySpec = null;
        if (querySpec == null && targetResourceType != null) {
            querySpec = getQuerySpec(targetResourceType);
        }
        if (querySpec == null && targetResourceClass != null) {
            querySpec = getQuerySpec(targetResourceClass);
        }
        if (querySpec == null) {
            querySpec = new QuerySpec(targetResourceClass, targetResourceType);
            if (targetResourceClass != null) {
                relatedSpecs.put(targetResourceClass, querySpec);
            }
            if (targetResourceType != null) {
                relatedSpecs.put(targetResourceType, querySpec);
            }
        }
        querySpec.setPaging(pagingSpec);
        return querySpec;
    }

The resource class is identical so this function returns QuerySpec of Test instead of TestRelation

What would be the best way to handle this case? Couldn't we get rid of this part of code:
if (querySpec == null && targetResourceType != null) {
            querySpec = getQuerySpec(targetResourceType);
        }
        if (querySpec == null && targetResourceClass != null) {
            querySpec = getQuerySpec(targetResourceClass);
        }

Best,
JB

remmeier:
where exactly does the problem happen? do you have a complete stack trace? In general the resourceType should take precedence over resourceClass. Crnk internally it should already always make use of the former. Application code has to do the same if it does similar things. Maybe some further checks could be added to detect issues in this area.

jbmolle:
Hi,
I'm actually having 2 dynamic objects: joboffers and companies. I would like to list joboffers and include companies in the api call so I do GET joboffers?include=companies.

Here is a paste of the part of the stack trace that shows the problem:
at io.crnk.core.engine.internal.repository.ResourceRepositoryAdapterImpl$3.invoke(ResourceRepositoryAdapterImpl.java:115)
[...]

PR #656, opened 2019-12-16
QuerySpecs creation with targetResourceType or targetResource Class

jbmolle:
Hi @remmeier 
This PR should solve #636 
I've split relatedSpecs between resourceType and resourceClass as you proposed.
I've used a Set to return the relatedSpecs in order to remove duplicates Specs (those who are both related to a resourceType and a resourceClass).
Let me know if it is what you were looking for.
Thx
JB

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/queryspec/QuerySpec.java
+++ b/crnk-core/src/main/java/io/crnk/core/queryspec/QuerySpec.java
@@ -273,13 +273,13 @@ public class QuerySpec {
     }
 
     public void setNestedSpecs(Collection<QuerySpec> specs) {
-    	this.typeRelatedSpecs.clear();;
+    	this.typeRelatedSpecs.clear();
         this.classRelatedSpecs.clear();
         for (QuerySpec spec : specs) {
-            if (spec.getResourceClass() != null) {
-                classRelatedSpecs.put(spec.getResourceClass(), spec);
-            } else {
+	    if (spec.getResourceType() != null) {
                 typeRelatedSpecs.put(spec.getResourceType(), spec);
+            } else {
+                classRelatedSpecs.put(spec.getResourceClass(), spec);
             }
         }
     }
--- a/crnk-core/src/main/java/io/crnk/core/queryspec/QuerySpec.java
+++ b/crnk-core/src/main/java/io/crnk/core/queryspec/QuerySpec.java
@@ -278,7 +278,8 @@ public class QuerySpec {
         for (QuerySpec spec : specs) {
 	    if (spec.getResourceType() != null) {
                 typeRelatedSpecs.put(spec.getResourceType(), spec);
-            } else {
+            } 
+        if (spec.getResourceClass() != null) {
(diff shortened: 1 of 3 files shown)


jbmolle:
Hi,
I think that we shouldn't allow QuerySpec to have null resourceType.
On the second commit, in QuerySpec if the resourceType is null then I get it from the resourceClass through the JsonApiResource annotation.

We have actually 2 cases:
- Either there is one resource class for each resource type. So we can map the resource type to the resource class (through JsonApiResource annotation)
- Or there is one resource class for multiple resource types. For example if we are creating resources dynamically at runtime. Then when we create a QuerySpec, we must specify the resource type. Just giving the resource class won't be enough to know which resources it corresponds to.

Currently if we have multiple resource types for one resource class, the mappedBy relations won't work. The ForwardOppositeStrategy repository won't build the correct queryspecs and the application return an error 500.
This PR also solves this problem.

jbmolle:
Hi @remmeier 
Did you have time to review this PR?
Let me know if I need to change something.
Thanks!

crnk-core/src/main/java/io/crnk/core/queryspec/QuerySpec.java
```diff
@@ -261,16 +266,20 @@ public void setIncludedRelations(List<IncludeRelationSpec> includedRelations) {
...
+    	allRelatedSpecs.addAll(classRelatedSpecs.values());
+        return Collections.unmodifiableCollection(allRelatedSpecs);
     }
 
     public void setNestedSpecs(Collection<QuerySpec> specs) {
-        this.relatedSpecs.clear();
+    	this.typeRelatedSpecs.clear();;
+        this.classRelatedSpecs.clear();
         for (QuerySpec spec : specs) {
             if (spec.getResourceClass() != null) {
```
remmeier: I suspect here the getResourceType() should have priority?

jbmolle: Yes you're right. I've just updated the pull request to reflect the changes

remmeier:
does .`/gradlew build` work for you? it seems that the travis integration is not working just now. Have to take a look at that. If so we could merge it.

jbmolle:
Now it should work.
Actually in setNestedSpecs we shouldn't do first resourceType and then resourceClass. We add it to both typeRelatedSpecs and classRelatedSpecs.
When doing getNestedSpecs, we send back a Set so QuerySpecs both in typeRelatedSpecs and classRelatedSpecs won't be sent twice.

The problem came from setNestedSpecWithClass() test:
we do Assert.assertSame(relatedSpec, spec.getQuerySpec(Task.class)); but in setNestedSpecs we were adding it only to typeRelatedSpecs so it wasn't in classRelatedSpecs.
The test Assert.assertSame(relatedSpec, spec.getQuerySpec("tasks")); was working.
Sorry about that.

remmeier:
perfect thx! a crnk 3.2 should be available within the coming days in case you are on latest.

Issue #671, opened 2020-01-17
LinksInformation links as object instead of string


jbmolle:
Hi @remmeier 
In JSON:API version 1.1,  links can have href, rel, anchor, params, describedby, meta.
What would you think of changing SelfLinksInformation so it would return a Link instead of a String?

The link interface could be something like this:
public interface Link {

	String getHref();

	void setHref(String href);

	@JsonInclude(JsonInclude.Include.NON_NULL)
	String getAnchor();

	void setAnchor(String anchor);

	@JsonInclude(JsonInclude.Include.NON_NULL)
	LinkParams getParams();

	void setParams(LinkParams params);

	@JsonInclude(JsonInclude.Include.NON_NULL)
	String getDescribedby();

	void setDescribedby(String describedby);

	@JsonInclude(JsonInclude.Include.NON_NULL)
	LinkMeta getMeta();

	void setMeta(LinkMeta meta);
}

So there is the describedby element in link which can point to an OpenApi or JsonSchema reference.

I could work on it if you're interested, let me know.

Best,
JB

remmeier:
would make sense. For relationship, self and paging link, we probably should stick to the short string-based representation. But custom links should be able to do this.

There already is some support for it. See `ClientType.ClientType.OBJECT_LINKS` and `CrnkProperties.SERIALIZE_LINKS_AS_OBJECTS`. But this would fully enforce object-based links more of a hidden feature just now. An explicit link object as you mention would be perfect I think.

remmeier:
we have this now, so closing it. Thank you!

PR #673, opened 2020-01-21
Links model

jbmolle:
Hi @remmeier 

This PR should solve #671 

Links will be serialized as objects if serializeLinksAsObjects property is true or if a link contains more than just href.

For now we can add href, describedby, anchor, rel as strings. Params and meta interfaces have been created but not used for now.

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/resource/links/DefaultLink.java
+++ b/crnk-core/src/main/java/io/crnk/core/resource/links/DefaultLink.java
@@ -1,0 +1,92 @@
+package io.crnk.core.resource.links;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+public class DefaultLink implements Link {
+
+	@JsonInclude(JsonInclude.Include.NON_NULL)
+	private String href;
+
+	@JsonInclude(JsonInclude.Include.NON_NULL)
+	private String rel;
+
+	@JsonInclude(JsonInclude.Include.NON_NULL)
+	private String anchor;
+
+	@JsonInclude(JsonInclude.Include.NON_NULL)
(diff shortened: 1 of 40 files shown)


remmeier:
will have a look tomorrow, thx!

crnk-core/src/main/java/io/crnk/core/resource/links/DefaultLink.java
```diff
@@ -0,0 +1,92 @@
+package io.crnk.core.resource.links;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+public class DefaultLink implements Link {
```
remmeier: currently wondering whether we should have an interface and default implementation or not. For all the other objects so far we did not have one. But unsure.

remmeier:
looks great! Not 100% backward-compatible, but easy to fix. so that point seems ok-ish to be compared to introducing a second set of objects.

crnk-core/src/main/java/io/crnk/core/resource/links/DefaultLink.java
jbmolle: Right. I've created an interface to follow the pattern of ResourceList and DefaultResourceList.
If also you have time to look at PR #656 that would be great. I need it for a project and currently I'm building a jar file of crnk-core instead of importing through Gradle.

jbmolle:
I tried to the generate links from strings so there isn't much change in code and the backward compatibility wouldn't be too bad. But I didn't manage to make it fully backward compatible :(

remmeier:
let us leave it like that, looks very good. thank you! was only holidays, but now also had a look at the other PR. Left one small comment.

PR #675, opened 2020-01-24
Make @crnk/ngrx-json compatible with ngrx v7+

gianinbasler:
Replaced .ofType() with .pipe(ofType()).
Updated ngrx dependencies to v7.4.0.


crnk-client-angular-ngrx/operations/crnk.operations.effects.ts
```diff
@@ -62,9 +62,9 @@ export class OperationsEffects {
 		'Accept': 'application/json-patch+json'
 	});
 
-	@Effect() applyResources$ = this.actions$
-		.ofType(NgrxJsonApiActionTypes.API_APPLY_INIT)
-		.withLatestFrom(this.store, (action, state: any) => {
+	@Effect() applyResources$ = this.actions$.pipe(
```
remmeier: is this sufficient or are there more places left in the project?

remmeier: ping @gianinbasler

gianinbasler: This is the only place I found in the project which is not compatible with ngrx v7+.

Issue #702, opened 2020-03-06
Filtering on a property value inside a two-level (or more) collection throws an exception


duncanportelli:
When filtering on a nested attribute where two or more of its parents is a Collection,

`filter[nestedbeans.subnestedbeans.property]=value`

an exception is being thrown:

`Cannot find an getter for io.crnk.core.resource.list.DefaultResourceList.property`

I believe this case is not handled in `PropertyUtils.getProperty(Object bean, List<String> propertyPath)`.

PR #703, opened 2020-03-06
Get property where attribute path contains more than one iterable

duncanportelli:
Fixed issue #702 where getting a property of a bean where more than one of its parents is a collection.


duncanportelli:
@remmeier , can you please review?

remmeier:
who is the user of that method? it may should be entirely deprected in favor of `BeanInformation?` which is used in most (but not quite all places). We may or may not have to apply a fix in that area as well.

duncanportelli:
@remmeier,

Method is used to retrieve values from objects based on the attribute path of the filter spec. This is done when calling querySpec.apply() to match the objects according to the filter

remmeier:
thx! let us merge like that then and maybe cleanup the BeanInformation at a later time. All the test cases are now in place to make sure things keep working how they are now.

PR #704, opened 2020-03-17
Add Prioritizing to default ExceptionMapper

alexander-matthiesen:
- Added different Priority-Levels to Prioritizable-Interface (1=HIGHEST_PRIORITY, 10=HIGH_PRIORITY, 50=MEDIUM_PRIORITY, 100=LOW_PRIORITY)
- Added Prioritizable to CrnkExceptionMapper and TimeoutExceptionMapper with MEDIUM_PRIORITY
- Moved adding default exceptionmapper before prioritizing them in ExceptionMapperRegistryBuilder

### DIFF
--- a/crnk-client/src/main/java/io/crnk/client/CrnkClient.java
+++ b/crnk-client/src/main/java/io/crnk/client/CrnkClient.java
@@ -135,6 +135,8 @@ public class CrnkClient {
 
     private ClientProxyFactory configuredProxyFactory;
 
+    private boolean orderExceptionMapperAfterClass = true;
+
     public enum ClientType {
         SIMPLE_lINKS,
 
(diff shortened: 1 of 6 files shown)


alexander-matthiesen:
@remmeier This might not work because of the logic how the ExceptionMappers are searched. It is currently by how far the ExceptionClasses are laying together. Do you have anything in mind how we can change that? 
We could just go with the priority or have a mix, but I guess the logic is very hard if we create a mix between class-depth and priority.

remmeier:
I think there is already a test case something in the area of the ExceptionMapperRegistry. It can be used to verify the changes. With negative priority it should already work today. But having an explicit priority seems like a good idea and much better than having to rely on negative ones.

alexander-matthiesen:
Unfortunately a negative Prio doesn't work currently.
So should we use just the Prio instead of the Exception-Depth-Search? Or do you have another suggestion?

remmeier:
that would be very breaking. In general I think the current mechanism going through the type hierarchy is fine enough. And then at second place comes the priority. So if for the same class multiple mappers exists, the one with the highest priority will win. `ExceptionMapperRegistryTest` already checks such use cases.

remmeier:
but maybe there is potential to have it more customizable. Should maybe looked at based on concrete use cases.

alexander-matthiesen:
Yes, I think the problem is, that in our case we want all Exceptions over one ExceptionMapper. Do you think there is a way to make that configurable that you could say the following:
"If EXCEPTION_MAPPER_ORDER=PRIORITY, then use only the priority, otherwise (which would be default, use EXCEPTION_CLASS_HIERARCHY"?

I think if we can configure it somewhere, maybe in the client, that would be a good solution.

alexander-matthiesen:
Currently the Prioritizable returns 0 if the Object doesn't have a priority. From my understanding this is wrong because if an object doesn't have a priority, it should be handled at last right?
That's why currently the ModuleRegistryTest.testinformationBuilder() fails. 
Do you @remmeier have a suggestion how to proceed? I made the other class-hierarchy already configurable.

remmeier:
if you really want to replace everything built-in and have just one custom one, maybe let us make it first configurable to skip the default implementations? Just now I think there is not yet a need for multiple lookup strategies to keep the code base simple. We just have to make sure  that we don't enable what we don't need?

remmeier:
for this I think we should remove the default registration in ExceptionMapperRegistryBuilder. Rather move those default exception mappers into a new Module similar to the `CoreModule` and register it in `CrnkBoot`. `CrnkBoot` would then offer a boolean property to skip registrating that new `DefaultExceptionModule`. Would this make sense?

alexander-matthiesen:
Yes, this would definetly make sense! My approach is maybe a little bit too special and complex. I like your idea! 
Do you implement that or what was your plan?

remmeier:
If necessary, feel free to create a PR for CrnkBoot

Issue #707, opened 2020-03-25
JAX-RS integration and webPathPrefix


lodrantl:
I have a JAX-RS service (on Wildfly) and am trying to include Crnk. It has an ApplicationPath of /myapp, but I want Crnk to reside at /myapp/api.

When you set the `webPrefixPath` using `feature.getBoot().setWebPathPrefix("/api");` nothing happens.

In the CrnkFeature.java we have:
```
    public String getWebPathPrefix() {
        return UrlUtils.removeLeadingSlash(boot.getWebPathPrefix());
    }
```
which is used in JaxrsRequestContext.java:
```
	private String buildPath(UriInfo uriInfo) {
		String basePath = uriInfo.getPath();
		String webPathPrefix = feature.getWebPathPrefix();
		String path;
		if (webPathPrefix != null && basePath.startsWith(webPathPrefix)) {
			path = basePath.substring(webPathPrefix.length());
		} else {
			path = basePath;
		}
		return path == null || path.isEmpty() ? "/" : path;
	}
```
There are two problems here:

1. `basePath.startsWith(webPathPrefix)` is never true because basePath starts with a slash which is explicitly removed from webPathPrefix.
2. Even if you keep the leading slash that would mean the Crnk repositories are served both without prefix and with prefix which I do not think is correct.

Am I doing something wrong or is this an actual problem of the implementation.
Thanks for your help guys.

remmeier:
unfortunately different frameworks work a bit differently in the area. So at some point we started just ignoring preceding slashes. But not completely as it seems. Things should work well if you drop the prefix?`meaning just `api`  rather than `/api`? But this should change.

lodrantl:
No, the problem is that basePath received from uriInfo.getPath() contains a leading slash, while the webPathPrefix will never contain it.

Problem 2 still remains, would modifying CrnkFilter to only pass requests matching the webpathPrefix work?

And also I found a third problem. The basePath returned at
https://github.com/crnk-project/crnk-framework/blob/44718c0edc98d0fb20af35fbd46cb071003b04f8/crnk-setup/crnk-setup-rs/src/main/java/io/crnk/rs/JaxrsRequestContext.java#L77
does not contain the webpathPrefix so the urls in links point to the wrong targets. I can submit a PR with my proposed changes if you want.

remmeier:
PR would be great. Given that one has to be careful with different frameworks, we should focus on changes to `JaxrsRequestContext`, but try to avoid other ones.

[...]

PR #711, opened 2020-04-09
JAX-RS integrations webpath prefix handling

lodrantl:
Goes with #707.

I kept all the changes inside the jax-rs-setup. Tried to add some tests too, although the CrnkFilterTest ones seem kinda bad. The biggest problem is the leading/trailing slash handling. The uriInfo.getPath() returns a leading slash on Wildfly and no leading slash on Glassfish.

### DIFF
--- a/crnk-setup/crnk-setup-rs/src/main/java/io/crnk/rs/CrnkFilter.java
+++ b/crnk-setup/crnk-setup-rs/src/main/java/io/crnk/rs/CrnkFilter.java
@@ -1,6 +1,7 @@
 package io.crnk.rs;
 
 import io.crnk.core.engine.dispatcher.RequestDispatcher;
+import io.crnk.core.engine.internal.utils.UrlUtils;
 import io.crnk.rs.type.JsonApiMediaType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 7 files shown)


crnk-setup/crnk-setup-rs/src/main/java/io/crnk/rs/CrnkFilter.java
```diff
@@ -42,6 +42,12 @@ public CrnkFilter(CrnkFeature feature) {
 
 	@Override
 	public void filter(ContainerRequestContext requestContext) {
+		if (feature.getWebPathPrefix() != null) {
+			String path = UrlUtils.removeLeadingSlash(requestContext.getUriInfo().getPath());
+			if (!path.startsWith(UrlUtils.addTrailingSlash(feature.getWebPathPrefix()))) {
```
remmeier, line 47: I guess it would make sense to add that trailing slash to `getWebPathPrefix`? Currently it does not. But there is no use case for that.

remmeier: it would ease the check here

remmeier:
thank you!  I'm still a bit conflicted about those leading and trailing slashes. but still leaning towards not having them. Potentially the code base should also start using typed-objects for Path and URL rather than work with strings. That would avoid having to deal with the slashes in the first place and give more flexiblity in terms of what to serve or not.

Issue #712, opened 2020-04-11
Infinite Loop in PathBuilder (Version 3.1.20191113192440)


TeamFinal:
Hello,
depending on a race Condition I get stuck in an infinite Loop in the class PathBuilder (Version 3.1.20191113192440).
The problem started to appear once I decided to use a 3 tier ressource inheritance.

The problem is located in the following Method on lines 221-223
```java
213	private RegistryEntry findSelfOrSubtypeByField(RegistryEntry entry, String fieldName) {
214		if (entry.getResourceInformation().findFieldByName(fieldName) == null) {
215			// consider introducing RegistyEntry.getSubTypes in the future
216			for (RegistryEntry someEntry : resourceRegistry.getEntries()) {
217				ResourceInformation resourceInformation = someEntry.getResourceInformation();
218				ResourceField field = resourceInformation.findFieldByName(fieldName);
219				if (field != null && someEntry.isParent(entry)) {
220					RegistryEntry parentEntry = someEntry.getParentRegistryEntry();
221					while (parentEntry != null && parentEntry.getResourceInformation().findFieldByName(fieldName) != null) {
222						someEntry = parentEntry;
223					}
224					return someEntry;
225				}
226			}
227		}
228		return entry; // may not contain the field in case of errorous request
229	}
```
If the loop is entered it never can be left because parentEntry doesn't change and therefore the condition will not change.
I think there is the following line missing.
```java
parentEntry = someEntry.getParentRegistryEntry();
```
Happy Easter
Andreas

remmeier:
indeed, thank you. should I create a PR? or do you have time? do you have by chance an example model at hand to add as test case?

TeamFinal:
I'd prefer you to make the Pull Request, because I'm a git newbie and I don't want to do anything wrong.
Unfortunately I don't have an example model at hand that suits as test case.

remmeier:
a fix is in place, but test case msut still be added

PR #713, opened 2020-04-14
Infinite Loop in PathBuilder.findSelfOrSubtypeByField removed

fjf2002:
Hopefully solves #712


remmeier:
thank you! but let us keep the ticket open till also a test case is added

Issue #740, opened 2020-06-02
allow typescript generation without crnk library dependency


remmeier:
for plainjson format, this is already working. Should support the same for pure JSON:API,

PR #741, opened 2020-06-02
remove crnk npm dependency from TS generator #740

remmeier:
Typescript can now be generated standalone without depending on the deprecated crnk npm library.  Instead a crnk.ts file with some  common base classes  is generated next to all the resource classes. Such a setup was already in place  for the slightly simplified plain-json format. This change brings the same setup to the regular JSON:API format. A dependency to ngrx-json-api is still remaining and can also be made optional in the future.


remmeier:
merged already

PR #743, opened 2020-06-02
feat(crnk-core): introduced JacksonPropertyNameResolver abstraction

vicmosin:


crnk-core/src/main/java/io/crnk/core/engine/internal/jackson/LinksInformationSerializer.java
```diff
@@ -34,13 +38,15 @@ public void serialize(LinksInformation value, JsonGenerator gen, SerializerProvi
 
 		for (String attrName : info.getAttributeNames()) {
 			BeanAttributeInformation attribute = info.getAttribute(attrName);
```
remmeier, line 40: the setup can be slightly simplified: BeanAttributeInformation already have a getJsonName that can be used here. So the LinksInformationSerializer does not need the propertyNameResolver, but can work with those information objects.
It just has not been used so far.

But your introduced JacksonPropertyNameResolver still  is a good idea. Probably best to integrate it with BeanAttributeInformation/BeanInformation. There is one not so nice thing tough. BeanInformation is fully static so far. So if we start making it customizable, that customization has to be static as well. For now  that seems ok-ish. Maybe in the future we have a BeanInformationFactory that is available from CrnkBoot to avoid the static/global JVM-specific things. Maybe as a start we could create that BeanInformationFactory. There would be a singleton default instance BeanInformationFactory.DEFAULT or something like that. BeanInformatioan.of(...) would make use of that singleton. And the BeanInformationFActory would have a registerNameResolver method.

crnk-core/src/main/java/io/crnk/core/engine/internal/jackson/JacksonPropertyNameResolver.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.crnk.core.engine.information.bean.BeanAttributeInformation;
+
+import java.util.Optional;
+
+/**
+ * Holds logic of retrieving property name based on {@link JsonProperty} value
+ */
+public class JacksonPropertyNameResolver {
```
remmeier, line 11: maybe let's also introduce a NameResolver interface (for that BeanInformationFactory brought up below). We could maybe slightly shortcut the name to JacksonNameResolver. It will allow to come up with other naming resolvers,  e.g. if somebody prefers x-y-z over camel case xYZ.

crnk-core/src/main/java/io/crnk/core/engine/internal/jackson/JacksonResourceFieldInformationProvider.java
```diff
@@ -97,9 +97,9 @@ else if (JsonInclude.Include.NON_EMPTY.equals(value)) {
 
 	@Override
 	public Optional<String> getJsonName(BeanAttributeInformation attributeDesc) {
```
remmeier, line 99: not sure, but maybe this method gets obsoleted entirely by the (Property)NamingResolver

vicmosin: Not sure about this though.. issue is that `JacksonResourceFieldInformationProvider` holds not only json-property name resolver but also jackson-naming-strategy related thing.. moving that path to `BeanAttributeInformation` would require it to have an instance of `ObjectMapper` or `BeanInformationFactory.ObjectMapper`. Are you ok with this?

remmeier: how do you mean ` jackson-naming-strategy related thing.. `?

Currently `ResourceInformationProviderBase` looks like: 

```
protected String getJsonName(BeanAttributeInformation attributeDesc, ResourceFieldType fieldType) {
        if (fieldType == ResourceFieldType.ID && enforceIdName) {
            // referred to as id even if named differently in Java
            return "id";
        }
        for (ResourceFieldInformationProvider fieldInformationProvider : resourceFieldInformationProviders) {
            Optional<String> jsonName = fieldInformationProvider.getJsonName(attributeDesc);
            if (jsonName.isPresent()) {
                return jsonName.get();
            }
        }
        return attributeDesc.getName();
    }
```

Here it could rather make use of `attributeDesc.getJsonName`. Then the `JacksonResourceFieldInformationProvider.getJsonName` should not be necessary I think. Was hoping for the `BeanFactory` to be independent of `ObjectMapper`. But maybe a missed something.

vicmosin: It doesn't work this way unfortunately.. `JacksonResourceFieldInformationProvider` holds information about the current jackson naming strategy and therefor removing `JacksonResourceFieldInformationProvider.getJsonName` from snippet you posted would actually cause that crnk will loose its ability to respect jackson naming strategy property... meaning each `ResourceInformationProviderBase. getJsonName` would return either property default name (in case no JacksonProperty annotation used) or the value from that annotation (calculated by new JacksonNameResolver). Which means naming strategy will be totally ignored

remmeier: maybe let's focus on the comment above https://github.com/crnk-project/crnk-framework/pull/743#discussion_r434329831, currently we may not speak about the same things :-)

Issue #695, opened 2020-02-14
crnk-gen-openapi generated yaml file type, request and response relationship field in wrong palce


jameshwh:
There are two issues found for the generated openapi.yaml file:
### Type's first letter is in upper case (expect lower case)
I generated an openapi.yml for my project, but the "type" is not correct, for example, the first character is in upper case instead of "projects" as defined in the Project.java `(@JsonApiResource(type = "projects"))`
If we set "Projects" as type and make a request, the response is "Repository for a resource not found: Projects", is it a known issue or is there anything i need to configure to make it work?
I found the root cause of the issue is the genereated enum type should be "projects" instead of "Projects" as you see below, below is a snippet text from https://github.com/crnk-project/crnk-framework/blob/master/crnk-gen/crnk-gen-openapi/src/test/resources/gold/complex.yaml
```
ProjectsPostResourceReference:
      properties:
        id:
          $ref: '#/components/schemas/ProjectsIdResourceAttribute'
        type:
          description: The JSON:API resource type (Projects)
          enum:
          - Projects
          type: string
```

### Relationship fields are in attribute section in request and response
All the relationships elements are in attributes field, see below, tasks should be in relationships {} block, but it is in attribute block:
```
{
   "id":0,
   "type":"Projects",
   "attributes":{
      "data":{
         "customData":{
            "property1":"string",
            "property2":"string"
         },
         "data":"string",
         "due":"2020-02-13T21:19:06Z",
         "image":"string",
         "keywords":[
            "string"
         ]
      },
      "description":"string",
      "name":"string",
      "task":{
         "id":0,
         "type":"Tasks"
      },
      "tasks":[
         {
            "id":0,
            "type":"Tasks"
         }
      ]
   }
```

jameshwh:
@macleondbroad-wf

PR #744, opened 2020-06-03
Fix OpenAPI generator type field and relationships issues

peter-plochan:
* Generate OpenAPI resource types with correct letters case
* Fix schemas containing or defining resource attributes and relationships
  * relationship fields separated from attributes
  * only valid (containing at least one property) relationships and attributes schemas are generated
  * enhanced relationship inner structures
  * links (with explicitly defined self link) in separate schema

Fixes https://github.com/crnk-project/crnk-framework/issues/695

### DIFF
--- a/crnk-gen/crnk-gen-openapi/src/main/java/io/crnk/gen/openapi/internal/schemas/TypeSchema.java
+++ b/crnk-gen/crnk-gen-openapi/src/main/java/io/crnk/gen/openapi/internal/schemas/TypeSchema.java
@@ -12,8 +12,8 @@ public class TypeSchema extends AbstractSchemaGenerator {
 
   public Schema schema() {
     Schema typeSchema = new StringSchema()
-        .description("The JSON:API resource type (" + metaResource.getName() + ")");
-    typeSchema.addEnumItemObject(metaResource.getName());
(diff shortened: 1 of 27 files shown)


peter-plochan:
@remmeier @macleodbroad-wf

crnk-gen/crnk-gen-openapi/src/main/java/io/crnk/gen/openapi/internal/schemas/TypeSchema.java
```diff
@@ -13,7 +13,7 @@
   public Schema schema() {
     Schema typeSchema = new StringSchema()
         .description("The JSON:API resource type (" + metaResource.getName() + ")");
```
macleodbroad-wf, line 15: You may want to update the description to match.

macleodbroad-wf:
I checked out the generated files in the Swagger Editor (https://editor.swagger.io/) and reviewed the code. One comment on updating the generated description.

Looks good!

crnk-gen/crnk-gen-openapi/src/main/java/io/crnk/gen/openapi/internal/schemas/TypeSchema.java
peter-plochan: Fixed

PR #745, opened 2020-06-05
fix: respect JsonProperty annotation while generating links

### LINKED ISSUE #736: Version 3.2 ignores custom property names at Links object
After migration (from 3.1) to 3.2 custom property names in custom `LinksInformation` objects can't be used anymore.

Since the same logic already used in `JacksonResourceFieldInformationProvider.getJsonName()` it would be nice to use some abstraction (such as `JacksonPropertyNameResolver`) to incapsulate `@PropertyName` -> resulted json name conversion. 
@remmeier Do you have maybe something on your mind?

vicmosin:
Resolves #736


remmeier:
thank you!

Issue #761, opened 2020-07-13
Implementing BulkResourceRepository for a resource breaks the operations module for that resource


primitive-type:
Implementing the `BulkResourceRepository` interface for a particular resource breaks the operations module for that resource. After implementing `BulkResourceRepository` for a resource and then trying to update it via the operations module, the following error message is generated:

`experimental bulk support is currently limited to POST and DELETE`

When the `BulkResourceRepository` implementation is reverted, the operations module begins working again for the affected resources.

Using crnk version: 3.2.20200419165537

PR #762, opened 2020-07-13
Enable PATCH requests to BulkResourceRepository implementations via operations module

primitive-type:
This begins to address #761 but I am very open to any guidance because I am not sure this handles the issue completely. It does seem that support for `PATCH` requests here is already implemented or expected, and the check for `POST` and `DELETE` requests specifically had just not been updated to reflect that. Let me know if I am mistaken or going down the wrong path here. Happy to write tests as well if this is sufficient.

### DIFF
--- a/crnk-operations/src/main/java/io/crnk/operations/server/OperationsModule.java
+++ b/crnk-operations/src/main/java/io/crnk/operations/server/OperationsModule.java
@@ -263,8 +263,8 @@ public class OperationsModule implements Module {
             JsonPath repositoryPath = new ResourcePath(rootEntry, null);
 
             PreconditionUtil.assertTrue(
-                method.equals(HttpMethod.POST.toString()) || method.equals(HttpMethod.DELETE.toString()) ,
-                "experimental bulk support is currently limited to POST and DELETE"
+                method.equals(HttpMethod.POST.toString()) || method.equals(HttpMethod.DELETE.toString()) || method.equals(HttpMethod.PATCH.toString()) ,
+                "experimental bulk support is currently limited to POST, PATCH and DELETE"
             );
 
             Document requestBody = new Document();


crnk-operations/src/main/java/io/crnk/operations/server/OperationsModule.java
```diff
@@ -263,8 +263,8 @@ private boolean bulkExecuteOperations(List<OrderedOperation> operations, Operati
             JsonPath repositoryPath = new ResourcePath(rootEntry, null);
 
             PreconditionUtil.assertTrue(
```
remmeier, line 265: I don't think it is that easy... But glad to be proof  wrong on this. There is a BulkClientTest where patch-related test could be added.

primitive-type: What I don't quite understand is why `BulkResourceRepository` conflicts with the operations module. Because bulk `PATCH` requests can work via either option, just not both simultaneously.

remmeier: the operations module is a bit more generic, it can take any number of changes. While the buk patch will "just" take multiple updates from a single type. Ideally the operations module will make use of bulk repositories where available. in what sense you mean "conflict"? 

just now that part of the code base is also not entirely in my mind. So I'm not quite sure where problem manifest. One issue is probably the URL handling. PATCH request assume /api/{resource}/{id} whereas in the bulk case it can only be /api/{resource} and all the IDs are then extracted from the body. Something to account for in `ResourcePatchController`

primitive-type: I described the issue a bit better here: https://github.com/crnk-project/crnk-framework/issues/761

Essentially, implementing `BulkResourceRepository` for a given resource effectively breaks operations module `PATCH` requests for that resource. But for the same resource when not implementing `BulkResourceRepository`, `PATCH` requests via the operations module work as expected.

remmeier: yes, let us  just remove that check. Add a test case and fix whatever comes up.

PR #772, opened 2020-07-31
Fix premature builder invocation in okhttp

challarao:
Listeners are not invoked as request is built before their registration.


remmeier:
thank you!

PR #784, opened 2020-09-07
Patch, Post resource relationships on non owner field

jbmolle:
Hi,

Currently it's not possible to Patch/Post a resource with a mappedBy relationship (where the field is not the owner of the relationship).
So that may be a problem because someone using the API may not have information on which relationship field is owner and which relationship field is owned.
This PR is a first attempt to solve that. It is a rework of #697 

Best,
JB

### DIFF
--- a/crnk-client/src/main/java/io/crnk/client/internal/ClientResourceUpsert.java
+++ b/crnk-client/src/main/java/io/crnk/client/internal/ClientResourceUpsert.java
@@ -2,11 +2,7 @@ package io.crnk.client.internal;
 
 import java.io.IOException;
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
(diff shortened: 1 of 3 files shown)


remmeier:
it goes a bit into a dangerous area. JSON:API/restful services have a rather strong semantic of resources and a single repositry being responsible for a resource. An update should touch it accordingly. This is reflected by the ownership. By allowing to update it from the opposite non-owning side, updates become much much more complex and can involve many repositories.

 there are  alternatives (orthogonal concerns, can also be combined).
- updating on the owner-side with multiple calls
- doing bulk/operations requests (there seems also progress on the JSON:APi spec side about this matter). 
- make use of the relationship endpoints

jbmolle:
Hi @remmeier 
Thanks for your feedback!
I understand your point. I could use the relationship endpoints but I thought it would be easier to update directly the document relationships with a PATCH request.
I think it follows the JSON:API spec version 1.1 under "Updating a Resource’s Relationships" paragraph. Any or all of a resource’s relationships MAY be included in the resource object included in a PATCH request. It doesn't talk about ownership in the paragraph.
The pull request doesn't try to update the attributes of the relationship.
So if you have companies and joboffers with companies having many joboffers and a joboffer belonging to one company. You should be able to update the company relationship { id: xxxx, type: "companies" } on the joboffer as well as update the joboffers relationship [ { id: xxx, type: "joboffers" } ] on the company.

But you're right, we can also use the relationship endpoints and I think the bulk/operations requests spec that's coming will also be very useful.

Thanks,
JB

remmeier:
I entirely see the point. Maybe it would be good to have it as add-on for a start. or to have an annotation on the relationship to allow this kind of propagation (similar to JPA allows to specify propagation behavior). But the default behavior should be simple I think, meaning only touching one repository.

PR #793, opened 2020-10-31
Modified DocumentMapper to use ServiceUrlProvider for generating document self links

### LINKED ISSUE #789: Generated selfs links on top level does not respect ServiceUrlProvider
The self link which is generated in DocumentMapper (line 146-148) does not take the ServiceUrlProvider in advance. So the resulting link uses the host and path from the request URI. Which could lead to wrong links when the system is behind a proxy / load balancer etc.

pekka-ia:
This allows for relative urls (by emptying crnk.domain-name) and also improves configurability in situations where the system runs behind a proxy.

Fixes #789

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/DocumentMapper.java
+++ b/crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/DocumentMapper.java
@@ -1,5 +1,7 @@
 package io.crnk.core.engine.internal.document.mapper;
 
+import io.crnk.core.engine.internal.utils.UrlUtils;
+import io.crnk.core.engine.url.ServiceUrlProvider;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -143,7 +145,8 @@ public class DocumentMapper {
 							linksInformation = selfLinksInformation;
 						}
 
-						JsonApiUrlBuilder.UrlParameterBuilder urlBuilder = new JsonApiUrlBuilder.UrlParameterBuilder(requestUri.toString());
+						String selfLinkUrl = UrlUtils.concat(queryAdapter.getResourceRegistry().getServiceUrlProvider().getUrl(), requestContext.getPath());
+						JsonApiUrlBuilder.UrlParameterBuilder urlBuilder = new JsonApiUrlBuilder.UrlParameterBuilder(selfLinkUrl);
 						urlBuilder.addQueryParameters(requestContext.getRequestParameters());
 						selfLinksInformation.setSelf(urlBuilder.toString());
 					}
--- a/crnk-core/src/test/java/io/crnk/core/engine/internal/document/mapper/DocumentMapperTest.java
+++ b/crnk-core/src/test/java/io/crnk/core/engine/internal/document/mapper/DocumentMapperTest.java
@@ -78,7 +78,7 @@ public class DocumentMapperTest extends AbstractDocumentMapperTest {
 		QueryAdapter adapter = createAdapter(Task.class);
 		Mockito.when(container.getRequestContextBase().getRequestUri()).thenReturn(URI.create("http://localhost/api/foo"));
 		Document document = mapper.toDocument(toResponse(task), adapter, mappingConfig).get();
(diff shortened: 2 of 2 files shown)


pekka-ia:
Taking another look at this, there was something odd with the behaviour in tests vs. behaviour in use.

pekka-ia:
I think I have a solid fix now. Instead of attempting to use ServiceUrlProvider, using `requestContext.getBaseUrl()` seems to yield exactly the same results.

When the below config is applied:

`crnk.domain-name: ""`

then

```
GET http://localhost:21120/api/incentive-programs?fields[incentive-programs]=name

{
    "data": [
        {
            "links": {
                "self": "/api/incentive-programs/8377ecb2-d516-4ee1-9fd3-259c54a9119f"
            }
            // rest of the object
        }
    ],
    "links": {
        "self": "/api/incentive-programs?fields[incentive-programs]=name"
    }
}
```

With an explicit domain set, it looks like this:

`crnk.domain-name: "https://www.nbc.com"`

```
GET http://localhost:21120/api/incentive-programs?fields[incentive-programs]=name

{
    "data": [
        {
            "links": {
                "self": "https://www.nbc.com/api/incentive-programs/a2081411-5961-4841-8570-6d538ad51f11"
            }
            // rest of the object
        }
    ],
    "links": {
        "self": "https://www.nbc.com/api/incentive-programs?fields[incentive-programs]=name"
    }
}
```

and with no domain-name configuration set at all, it still works as expected

```
GET http://localhost:21120/api/incentive-programs?fields[incentive-programs]=name
{
    "data": [
        {
            "links": {
                "self": "http://localhost:21120/api/incentive-programs/bf871ce6-8f25-45c9-ae69-71823bbef46b"
            }
            // rest of the object
        }
    ],
    "links": {
        "self": "http://localhost:21120/api/incentive-programs?fields[incentive-programs]=name"
    }
}
```

remmeier:
It is a bit a dangerous area for changes because of the variety of integrations (servlet, etc.). Where have you seen that `reqestUri != baseUrl + path`? Because that would be wrong and in need for fixing. There might be other places where reqestUri is used and also breaks.

pekka-ia:
Finally got back to this. Basically there is similar behaviour here:

https://github.com/crnk-project/crnk-framework/blob/master/crnk-core/src/main/java/io/crnk/core/engine/internal/registry/ResourceRegistryImpl.java#L117-L120

https://github.com/crnk-project/crnk-framework/blob/master/crnk-core/src/main/java/io/crnk/core/engine/internal/registry/ResourceRegistryImpl.java#L143-L147

where QueryContext is the source of the baseUrl.

vicmosin:
@remmeier  Can we please merge this change? I need self link to be generated based on ServiceUrlProvider as well

crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/DocumentMapper.java
```diff
@@ -143,7 +145,8 @@ private LinksInformation enrichSelfLink(LinksInformation linksInformation, Query
 							linksInformation = selfLinksInformation;
 						}
 
-						JsonApiUrlBuilder.UrlParameterBuilder urlBuilder = new JsonApiUrlBuilder.UrlParameterBuilder(requestUri.toString());
+						JsonApiUrlBuilder.UrlParameterBuilder urlBuilder = new JsonApiUrlBuilder.UrlParameterBuilder(UrlUtils.concat(queryAdapter.getQueryContext().getBaseUrl(),
```
remmeier, line 148: what isnwring with the request uri? maybe a fix should be applied there if necessary. currently the code is inconsitent, having null check for resourceuri, but does not make use it.

remmeier (changes requested):

vicmosin:
@pekka-ia are you gonna provide the changes @remmeier asked for?

pekka-ia:
> @pekka-ia are you gonna provide the changes @remmeier asked for?

I don't think I can currently do it. I don't have a good environment for doing Java development currently (this PR was opened while I was working for my former employer). :/

PR #801, opened 2021-01-20
Fixed exception message to correctly contain the applicable application version

pekka-ia:
I was writing some tests around this in one of my projects and ran into this issue. Example of the error before the fix as it is looks like this:

```
{
    "errors": [
        {
            "status": "400",
            "title": "BAD_REQUEST",
            "detail": "attribute versionBetween110And120 not available for version {}"
        }
    ]
}
```


remmeier:
thx!

PR #815, opened 2021-05-27
fix(core): made sure self links at document level use service-url-provider as base url

### LINKED ISSUE #789: Generated selfs links on top level does not respect ServiceUrlProvider

vicmosin:
Resolves #789


vicmosin:
@remmeier 
made my version of https://github.com/crnk-project/crnk-framework/pull/793

1. I saw you comment about wrong place for calculating the proper URL and imo the place you suggested is also isn't perfect. I find `HttpRequestContextBaseAdapter` to be bound to current HTTP request, therefor it would be confusing to put there ServiceUrlProvider-related URLs. `QueryContext`, on the other hand, fits perfectly as it holds not only HTTP stuff but crnk context about current query, so I extended it.
2. I could not find a good place for integration test so used spring-boot-example project to have the new test which checks the links without any additional mocking

remmeier (approved):

remmeier:
thx!

datagitlies:
is there an ETA for a new release with this fix @remmeier ?

PR #821, opened 2021-08-11
Update CrnkErrorController to be compatible with Spring Boot >= 2.3.X

### LINKED ISSUE #816: Crnk error handling is not compatible with Spring Boot 2.5
In particular `CrnkErrorController.errorToJsonApi()` calls `AbstractErrorController.getErrorAttributes()` which takes a `boolean` in 2.4 but in 2.5 expects an `ErrorAttributeOptions` resulting in the following stacktrace:

```
java.lang.NoSuchMethodError: 'java.util.Map io.crnk.spring.setup.boot.mvc.CrnkErrorController.getErrorAttributes(javax.servlet.http.HttpServletRequest, boolean)'
	at io.crnk.spring.setup.boot.mvc.CrnkErrorController.errorToJsonApi(CrnkErrorController.java:39) ~[crnk-setup-spring-boot2-3.4.20210509072026.jar:na]
	at io.crnk.spring.setup.boot.mvc.CrnkErrorController.error(CrnkErrorController.java:64) ~[crnk-setup-spring-boot2-3.4.20210509072026.jar:na]
	...
```

datagitlies:
The `getErrorAttributes` method that the `CrnkErrorController` was calling had been deprecated in Spring Boot 2.3 and then was removed in Spring Boot 2.5 - this pull request updates the controller to use the new / non-deprecated version of the `getErrorAttributes` method.

Closes #816

### DIFF
--- a/crnk-setup/crnk-setup-spring-boot2/build.gradle
+++ b/crnk-setup/crnk-setup-spring-boot2/build.gradle
@@ -29,6 +29,7 @@ dependencies {
 	compileOnly 'org.springframework.boot:spring-boot-starter-web'
 	compileOnly 'org.springframework.boot:spring-boot-configuration-processor'
 	compileOnly 'org.springframework.boot:spring-boot-starter-actuator'
+    compileOnly 'org.springframework.boot:spring-boot-starter-validation'
 	compileOnly 'org.springframework.security:spring-security-core'
 
 	compileOnly 'org.springframework:spring-orm'
@@ -49,6 +50,7 @@ dependencies {
 	testCompile 'org.springframework.boot:spring-boot-starter-web'
 	testCompile 'org.springframework.boot:spring-boot-starter-test'
 	testCompile 'org.springframework.boot:spring-boot-starter-actuator'
+	testCompile 'org.springframework.boot:spring-boot-starter-validation'
 
 	testCompile 'org.hibernate:hibernate-core'
 	testCompile 'org.hibernate:hibernate-entitymanager'
--- a/crnk-setup/crnk-setup-spring-boot2/build.gradle
+++ b/crnk-setup/crnk-setup-spring-boot2/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'java'
 apply plugin: 'io.spring.dependency-management'
 dependencyManagement {
 	imports {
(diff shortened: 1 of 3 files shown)


remmeier:
if server.error.include-message=always.is always reqired,.maybe it should be added to the spring configuration to enable it by default?

datagitlies:
> if server.error.include-message=always.is always reqired,.maybe it should be added to the spring configuration to enable it by default?

I only set the value to fix a broken test. I think crnk should be agnostic to this setting. The main thing is formatting the error per the JSON:API specification. The end user can then decide to include the message, or not, depending on their needs using the spring property. Crnk need not take a stance either way which I would think is best.

datagitlies:
@remmeier is there anything you'd like me to change?  Also, I'm curious if/when this and a few other fixes could be rolled up into a new release... I'm sort of stuck waiting on this pull request as well as #815

rohitb1985:
In which release of Crnk can we expect this to be fixed?

Incanus3:
I can confirm that this PR solves the issue. Tried in our project using the `--include-build ../crnk-framework` gradle option. The only problem I have when testing this way is that I can't use the `crnk-validations` module, because it fails with `class file for javax.validation.ValidatorFactory not found`, even though when running with crnk.io as maven dependency, I don't encounter this problem, but that is most probably a resolving issue and definitely not related to this issue. If you know, how to test this better though, please give me a hint.

remmeier:
ehatbteat is failing? if itnis an essential one, it erver.error.include-message=always should become a default. or else the flag docunented maybe. else let usbstash and merge then.

datagitlies:
@remmeier from my understanding, the `CrnkErrorController` that is "broken" here is not being used in any of the core pieces of the `crnk-framework`.  It only applies to applications using Spring Boot / Spring MVC. In my case, this error controller simply formats the response for anything other than my resource / relationship repository URLs.  Simple things like a 404 coming back from Spring's DispatcherServlet are then formatted according to the spec.  Very nice to have but not critical.  It looks like this issue is preventing several users of crnk from upgrading their Spring Boot version to `5.X` (myself included). Hence, we're all just wondering if/when this fix can be made and a new version of crnk released.

datagitlies:
@remmeier is there anything you'd like me to change? Also, I'm curious if/when this and a few other fixes could be rolled up into a new release... I'm sort of stuck waiting on this pull request as well as #815

Issue #790, opened 2020-10-15
JsonInclude NON_EMPTY on collection broken since 3.3


cgendreau:
`@JsonInclude(Include.NON_EMPTY)` behavior changed between 3.2 and 3.3.

In version 3.3, Crnk will ignore an array or collection field annotated with `@JsonInclude(Include.NON_EMPTY)` even if it contains something.

cgendreau:
I think the issue is related `io.crnk.core.engine.internal.document.mapper.ResourceMapper` :

```
protected boolean isDefaultNodeValue(JsonNode node) {
    return isNullNodeValue(node) || (node.isObject() || node.isArray()) && node.size() == 0
        || node.asText().isEmpty() || node.isNumber() && node.asDouble() == 0d;
}
```

Since `asText ` will return an empty string for container node I think it should check that it's a value node:

`node.isValueNode() && node.asText().isEmpty()`

If it sounds right I will send a PR.

PR #822, opened 2021-08-12
Fix JsonInclude NON_EMPTY on collection

cgendreau:
* Added JsonInclude Test
* Fixed JsonInclude.Non_Empty
* isDefaultNodeValue method fix

### DIFF
--- a/crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/ResourceMapper.java
+++ b/crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/ResourceMapper.java
@@ -186,7 +186,7 @@ public class ResourceMapper {
 
 	protected boolean isDefaultNodeValue(JsonNode node) {
 		return isNullNodeValue(node) || (node.isObject() || node.isArray()) && node.size() == 0
-				|| node.asText().isEmpty() || node.isNumber() && node.asDouble() == 0d;
+				|| node.isValueNode() && node.asText().isEmpty() || node.isNumber() && node.asDouble() == 0d;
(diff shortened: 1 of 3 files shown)


cgendreau:
Fix Issue #790

crnk-core/src/main/java/io/crnk/core/engine/internal/document/mapper/ResourceMapper.java
```diff
@@ -186,7 +186,7 @@ private boolean isNodeIncluded(ResourceField field, JsonNode node) {
 
 	protected boolean isDefaultNodeValue(JsonNode node) {
 		return isNullNodeValue(node) || (node.isObject() || node.isArray()) && node.size() == 0
-				|| node.asText().isEmpty() || node.isNumber() && node.asDouble() == 0d;
+				|| node.isValueNode() && node.asText().isEmpty() || node.isNumber() && node.asDouble() == 0d;
 	}
```
remmeier, line 190: is there possibly a more efficient way? I would expect that asText to be rather inefficient for non primitve fields.

Jkeyuk: I believe the intention of the author was to skip the ` node.asText().isEmpty()` for fields that were not primitive. As I believe `node.isValueNode()` will return false for JSON types that are arrays or objects.

remmeier: ok thx!

PR #853, opened 2022-09-07
Requested fields on super resource type are now taken into account.

### LINKED ISSUE #852: Requested fields using super resource type are not taken into account
When using request parameter field, if specifying a super type of the resource requested, the parameter is not taken into account.

Let's define two resources, SuperTask and SpecialTask, the latter inheriting the former, using a single repository per type hierarchy (as stated here : https://www.crnk.io/releases/stable/documentation/#_inheritance) :

`
@JsonApiResource(type = "superTasks", subTypes = SpecialTask.class)
public abstract class SuperTask {

	@JsonApiId
	private Long id;

	private String name;

	private String category;
}

@JsonApiResource(type = "specialTask", resourcePath = "superTasks")
public class SpecialTask extends SuperTask {

	private boolean recurring;

	private String end;
}
`
Getters and setters have been omitted for brevity.

Requesting path superTasks with request parameter :
* :heavy_check_mark:  `fields[specialTask]=name` returns json including only attribute `name`.
* :heavy_check_mark:  `fields[specialTask]=end` returns json including only attribute `end`.
* :heavy_check_mark:  `fields[specialTask]=name,end` returns json including only attributes `name` and `end`.
* :negative_squared_cross_mark:  `fields[superTasks]=name`returns json including all attributes (from SpecialTask and SuperTask).
* :negative_squared_cross_mark:  `fields[superTasks]=name` and `fields[specialTask]=end` returns json including only attribute `end`.

Expected responses :
* :heavy_check_mark:  `fields[specialTask]=name` returns json including only attribute `name`.
* :heavy_check_mark:  `fields[specialTask]=end` returns json including only attribute `end`.
* :heavy_check_mark:  `fields[specialTask]=name,end` returns json including only attributes `name` and `end`.
* :heavy_check_mark:  `fields[superTasks]=name` returns json including only attribute `name`.
* :heavy_check_mark:  `fields[superTasks]=name` and `fields[specialTask]=end` returns json including only attributes `name` and `end`.

At the moment, introducing a new type in the hierarchy (a sibling for SpecialTask) creates a breaking change for our customers, because we need them to specifiy fields for all sub-types.

lgringo:
* Update DocumentMapperUtil.getRequestedFields to read requested fields on super type.
* Add test case.

Resolve #852


lgringo:
Rebased with master

Issue #856, opened 2022-09-14
2 query parameters 'fields' on different resources but on the same set of fields cancel the last one


lgringo:
# Issue

**Given** a resource named `main` having two related resources of type `a` and `b`.
**When** requesting `main` using `fields[a]=label` and `fields[b]=label`
**Result** all fields of resource `b` are returned.
**Expected** only `label` must be returned.

# Example :

## Resources

```java

@JsonApiResource(type = "main")
public class MainResource {

	@JsonApiId
	private String id;

	@JsonApiRelation
	private RelatedResourceA relatedResourceA;

	@JsonApiRelation
	private RelatedResourceB relatedResourceB;
}

@JsonApiResource(type = "a")
public class RelatedResourceA {
	@JsonApiId
	private String id;

	private String label;

	private String otherField;
}

@JsonApiResource(type = "b")
public class RelatedResourceB {
	@JsonApiId
	private String id;

	private String label;

	private String somethingDifferent;
}

``` 
NB: Getters and setters are omitted for brevity.

## Requests

NB: requests are written with [httpie](https://github.com/httpie), for these examples, you just have to know that 'x==y' means add a request parameter x with value y.

### No fields :+1: 

* `http :8080/main/mainId include==relatedResourceA,relatedResourceB`

```json
{
    "data": {
        "id": "mainId",
        "links": " [...] "
        "relationships": {
            "relatedResourceA": {
                "data": {
                    "id": "mainId.A",
                    "type": "a"
                },
                "links": " [...] "
            },
            "relatedResourceB": {
                "data": {
                    "id": "mainId.B",
                    "type": "b"
                },
                "links": " [...] "
            }
        },
        "type": "main"
    },
    "included": [
        {
            "attributes": {
                "label": "Label from resource A",
                "otherField": "Other"
            },
            "id": "mainId.A",
            "links": " [...] ",
            "type": "a"
        },
        {
            "attributes": {
                "label": "Label from resource B",
                "somethingDifferent": "Other"
            },
            "id": "mainId.B",
            "links": " [...] ",
            "type": "b"
        }
    ],
    "links": " [...] "
}
```
All fields from resource A and resource B are returned.

### Only label for resource A (or resource B) :+1: 

* `http :8080/main/mainId include==relatedResourceA,relatedResourceB 'fields[a]==label'`

```json
{
...
    "included": [
        {
            "attributes": {
                "label": "Label from resource A"
            },
            "id": "mainId.A",
[...]

PR #857, opened 2022-09-14
Add resourceType to QuerySpec.equals (and hashcode)

lgringo:
Fix for #856 :
* Adding resourceType to QuerySpec.equals.
* Fix TU failing


lgringo:
Rebased with master

